PR #8, opened 2020-06-30
Add unit testing dependency and starting unit test.

imnoahcook:
**Context**
One of the issues in the repo right now is to add unit testing.

**Solution**
I added Junit4 instead of Junit5 because Junit5 is not compatible with java 8, which is defined in https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/blob/master/build.gradle. If you feel that we should be supporting Java8, let's have a conversation.

The current test is a constructor test, which isn't all that useful in itself but it shows that the testing dependency works and allows anybody who wants to write a unit test in the future to have a reference to go off of.

In addition, I added some elements to the `.gitignore` of things that I thought were best not to commit

**Test plan**
```sh
$ cd exporters/trace && gradle test

BUILD SUCCESSFUL in 4s
3 actionable tasks: 3 up-to-date
```

### DIFF
--- a/exporters/trace/build.gradle
+++ b/exporters/trace/build.gradle
@@ -4,5 +4,5 @@ dependencies {
     api(libraries.opentelemetry_api)
     api(libraries.opentelemetry_sdk)
     api(libraries.google_cloud_trace)
-		testImplementation('junit:junit:4.13')
+    testImplementation('junit:junit:4.13')
 }
--- a/exporters/trace/build.gradle
+++ b/exporters/trace/build.gradle
@@ -1,7 +1,12 @@
+plugins {
+    id 'java'
+}
+
 description = 'Cloud Trace Exporter for OpenTelemetry'
 
 dependencies {
     api(libraries.opentelemetry_api)
     api(libraries.opentelemetry_sdk)
     api(libraries.google_cloud_trace)
+		testImplementation('junit:junit:4.13')
 }
--- a/exporters/trace/build.gradle
+++ b/exporters/trace/build.gradle
@@ -4,5 +4,5 @@ dependencies {
     api(libraries.opentelemetry_api)
     api(libraries.opentelemetry_sdk)
     api(libraries.google_cloud_trace)
-    testImplementation('junit:junit:4.13')
+    api(libraries.junit)
 }
--- a/exporters/trace/build.gradle
+++ b/exporters/trace/build.gradle
@@ -4,5 +4,5 @@ dependencies {
     api(libraries.opentelemetry_api)
     api(libraries.opentelemetry_sdk)
     api(libraries.google_cloud_trace)
-    api(testLibraries.junit)
+    testImplementation(testLibraries.junit)
 }
--- a/exporters/trace/build.gradle
+++ b/exporters/trace/build.gradle
@@ -4,5 +4,5 @@ dependencies {
     api(libraries.opentelemetry_api)
     api(libraries.opentelemetry_sdk)
     api(libraries.google_cloud_trace)
-    api(libraries.junit)
+    api(testLibraries.junit)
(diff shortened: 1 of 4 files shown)


exporters/trace/build.gradle
```diff
@@ -1,7 +1,12 @@
+plugins {
+    id 'java'
```
nilebox, line 2: I think this can be moved to https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/blob/5ecabaab3b959a8e882bd2333691137e49d35bc5/build.gradle#L2, so that we only make this change once to be used by all projects in the repo.

exporters/trace/build.gradle
```diff
@@ -1,7 +1,12 @@
...
+}
+
 description = 'Cloud Trace Exporter for OpenTelemetry'
 
 dependencies {
     api(libraries.opentelemetry_api)
     api(libraries.opentelemetry_sdk)
     api(libraries.google_cloud_trace)
+		testImplementation('junit:junit:4.13')
```
nilebox, line 11: Indentation looks wrong

exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/TraceExporterTest.java
```diff
@@ -0,0 +1,23 @@
...
+import com.google.devtools.cloudtrace.v2.AttributeValue;
+import static org.junit.Assert.assertNotNull;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+import java.util.Map;
+import java.util.HashMap;
+import org.junit.Test;
+
+@RunWith(JUnit4.class)
+
```
nilebox, line 12: No need for blank line here

exporters/trace/build.gradle
nilebox: Are you sure we need the `java` plugin?
Looks like `java-library` plugin itself supports `testImplementation`, it's mentioned on https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_configurations_graph

imnoahcook: you're right. `gradle build` works without it.

nilebox: And `gradle test` as well?

exporters/trace/build.gradle
imnoahcook: mb. was using tabs and it looked the same to my editor

exporters/trace/build.gradle
```diff
@@ -4,4 +4,5 @@ dependencies {
     api(libraries.opentelemetry_api)
     api(libraries.opentelemetry_sdk)
     api(libraries.google_cloud_trace)
+    testImplementation('junit:junit:4.13')
```
nilebox, line 7: Sorry, missed this: Could you please define a constant in https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/blob/5ecabaab3b959a8e882bd2333691137e49d35bc5/build.gradle#L20 and refer to it here?
This will help us use the same version across the repository.

e.g. `testLibraries.junit`

imnoahcook: Changed and thanks for the advice. I'm still learning about how gradle works and this seems much cleaner :)

exporters/trace/build.gradle
```diff
@@ -4,4 +4,5 @@ dependencies {
     api(libraries.opentelemetry_api)
     api(libraries.opentelemetry_sdk)
     api(libraries.google_cloud_trace)
+    api(testLibraries.junit)
```
nilebox, line 7:
```suggestion
    testImplementation(testLibraries.junit)
```

PR #12, opened 2020-07-17
Create a TraceConfiguration file and unit tests

imnoahcook:
# Context
Opencensus had a `StackdriverTraceConfiguration` file. This made it easy for customers to use the exporter while taking information from environment variables like project ID automatically. In addition, it abstracted away the details of making your own TraceServiceClient, which was not very clear how to do.

# Solution
I created an abstract class which is generated using the @AutoValue annotation. This configuration should be able to be used to create a configuration, which will be able to be passed as an argument to the TraceExporter by customers.
The configuration contains many of the attributes important to customers, such as createTraceServiceStub.

In the future this will be integrated with the actual TraceExporter class for customers to make exporters.

# Test plan
```sh
$ gradle test

BUILD SUCCESSFUL in 1s
3 actionable tasks: 3 up-to-date
```

### DIFF
--- a/exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceConfiguration.java
+++ b/exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceConfiguration.java
@@ -29,8 +29,7 @@ public abstract class TraceConfiguration {
     @VisibleForTesting
     static final Duration DEFAULT_DEADLINE = Duration.ofSeconds(10, 0);
 
-    TraceConfiguration() {
-    }
+    TraceConfiguration() {}
 
     /**
      * Returns the {@link Credentials}.
--- a/exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceConfiguration.java
+++ b/exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceConfiguration.java
@@ -8,6 +8,7 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
 import com.google.devtools.cloudtrace.v2.AttributeValue;
+
 import java.util.Collections;
 import java.time.Duration;
 import java.util.LinkedHashMap;
@@ -17,7 +18,6 @@ import javax.annotation.concurrent.Immutable;
 
 /**
  * Configurations for {@link TraceConfiguration}.
- *
  */
 @AutoValue
 @Immutable
@@ -28,7 +28,8 @@ public abstract class TraceConfiguration {
     @VisibleForTesting
     static final Duration DEFAULT_DEADLINE = Duration.ofSeconds(10, 0);
 
-    TraceConfiguration() {}
+    TraceConfiguration() {
+    }
 
     /**
      * Returns the {@link Credentials}.
@@ -82,7 +83,6 @@ public abstract class TraceConfiguration {
 
     /**
      * Builder for {@link TraceConfiguration}.
-     *
      */
     @AutoValue.Builder
     public abstract static class Builder {
--- a/exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceConfiguration.java
+++ b/exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceConfiguration.java
@@ -9,12 +9,12 @@ import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
 import com.google.devtools.cloudtrace.v2.AttributeValue;
 
-import java.util.Collections;
-import java.time.Duration;
-import java.util.LinkedHashMap;
-import java.util.Map;
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.Immutable;
+import java.time.Duration;
+import java.util.Collections;
+import java.util.LinkedHashMap;
+import java.util.Map;
 
 /**
  * Configurations for {@link TraceConfiguration}.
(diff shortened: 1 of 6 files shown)


imnoahcook:
Marking as WIP until I integrate it with the TraceExporter file

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceExporter.java
```diff
@@ -18,6 +29,20 @@
...
+  static TraceExporter createWithCredentials(String projectId, Credentials credentials,
+      Map<String, AttributeValue> fixedAttributes, Duration deadline) throws IOException {
+    TraceServiceSettings.Builder builder = TraceServiceSettings.newBuilder()
+        .setCredentialsProvider(FixedCredentialsProvider.create(checkNotNull(credentials, "credentials")));
+    // We only use the batchWriteSpans API in this exporter.
+    builder.batchWriteSpansSettings().setSimpleTimeoutNoRetries(org.threeten.bp.Duration.ofMillis(deadline.toMillis()));
+    return new TraceExporter(projectId, TraceServiceClient.create(builder.build()), fixedAttributes);
+  }
+
   public TraceExporter(String projectId, TraceServiceClient traceServiceClient,
```
imnoahcook, line 46: considered making this ctor private. Didn't do it immediate because I would have to delete a unit test and wanted to get feedback first.

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceConfiguration.java
```diff
@@ -0,0 +1,169 @@
...
+
+/**
+ * Configurations for {@link TraceConfiguration}.
+ *
+ */
+@AutoValue
+@Immutable
+public abstract class TraceConfiguration {
+
+    private static final String DEFAULT_PROJECT_ID = Strings.nullToEmpty(ServiceOptions.getDefaultProjectId());
```
haileyzhou, line 26: The format here looks a bit weird and so do some other places. Would you please apply the google-java-format checker or (if you have) check why it's not working?

exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/TraceConfigurationTest.java
```diff
@@ -0,0 +1,119 @@
...
+    private static final String PROJECT_ID = "project";
+    private static final Duration ONE_MINUTE = Duration.ofSeconds(60, 0);
+    private static final Duration NEG_ONE_MINUTE = Duration.ofSeconds(-60, 0);
+
+    @Test
+    public void defaultConfiguration() {
+        TraceConfiguration configuration;
+        try {
+            configuration = TraceConfiguration.builder().build();
+        } catch (Exception e) {
```
haileyzhou, line 37: I am not sure if I understand the testing purpose here. Was this for testing whether an exception will be thrown when project_id is not set? If so, CMIW, I think we can probably just avoid mixing the case here as it seems to be covered by disallowEmptyProjectId()?

exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/TraceConfigurationTest.java
```diff
@@ -0,0 +1,119 @@
...
+        }
+        assertNull(configuration.getCredentials());
+        assertNotNull(configuration.getProjectId());
+        assertNull(configuration.getTraceServiceStub());
+        assertTrue(configuration.getFixedAttributes().isEmpty());
+        assertEquals(TraceConfiguration.DEFAULT_DEADLINE, configuration.getDeadline());
+    }
+
+    @Test
+    public void updateAll() {
```
haileyzhou, line 49: nit: if it's possible, could you please maybe update it with a clearer name. The current name doesn't seem very straightforward for someone lack of context.

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceExporter.java
haileyzhou: nit: Please consider adding a TODO if it doesn't come right after this change.

exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/TraceConfigurationTest.java
```diff
@@ -0,0 +1,119 @@
...
+    @Test
+    public void defaultConfiguration() {
+        TraceConfiguration configuration;
+        try {
+            configuration = TraceConfiguration.builder().build();
+        } catch (Exception e) {
+            // Some test hosts may not have cloud project ID set up.
+            configuration = TraceConfiguration.builder().setProjectId("test").build();
+        }
+        assertNull(configuration.getCredentials());
```
haileyzhou, line 41: Some best-practices tips:
 - Split tests into well-defined arrange, act, and assert blocks
 - Ensure that the blocks are visually distinct from one another via line breaks or comments

Please apply throughout :)

exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/TraceConfigurationTest.java
imnoahcook: I'll fix it by setting project ID explicitly.

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceExporter.java
imnoahcook: Made it private

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceConfiguration.java
imnoahcook: I applied it. It looks a bit weirder than before in my opinion, but it's following a consistent standard so I suppose its better.

imnoahcook:
I ran the google Java formatter on the Java files. It looks pretty weird for my taste, but it follows a consistent style at least.

haileyzhou:
LGTM

imnoahcook:
Update ran google java formatter again. For some reason it looks cleaner and better, thought it would be nice to commit the better looking version

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceExporter.java
```diff
@@ -1,25 +1,79 @@
...
 public class TraceExporter implements SpanExporter {
 
   private final TraceServiceClient traceServiceClient;
   private final ProjectName projectName;
   private final String projectId;
   private final Map<String, AttributeValue> fixedAttributes;
 
-  public TraceExporter(String projectId, TraceServiceClient traceServiceClient,
-      Map<String, AttributeValue> fixedAttributes) {
+  static TraceExporter createWithConfiguration(TraceConfiguration configuration)
```
nilebox, line 31: This and other method should be `public static`, otherwise end users won't be able to use it outside of this package.
```suggestion
  public static TraceExporter createWithConfiguration(TraceConfiguration configuration)
```

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceExporter.java
```diff
@@ -1,25 +1,79 @@
...
+          throws IOException {
+    String projectId = configuration.getProjectId();
+    TraceServiceStub stub = configuration.getTraceServiceStub();
+    Credentials credentials = configuration.getCredentials();
+
+    if (stub == null) {
+      return TraceExporter.createWithCredentials(
+              projectId,
+              credentials != null ? credentials : GoogleCredentials.getApplicationDefault(),
```
nilebox, line 40: `credentials` are only used here, so it probably makes sense to move the declaration of this variable inside the `if (stub == null) { ... }` block.

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceExporter.java
```diff
@@ -1,25 +1,79 @@
...
+              configuration.getFixedAttributes(),
+              configuration.getDeadline());
+    }
+    return TraceExporter.createWithStub(
+            projectId, TraceServiceClient.create(stub), configuration.getFixedAttributes());
+  }
+
+  static TraceExporter createWithStub(
```
nilebox, line 48: Stub refers to testing, and this method doesn't actually explicitly use stubs internally, so I suggest to rename it to

```suggestion
  static TraceExporter createWithClient(
```

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceExporter.java
```diff
@@ -1,25 +1,79 @@
...
+          Duration deadline)
+          throws IOException {
+    TraceServiceSettings.Builder builder =
+            TraceServiceSettings.newBuilder()
+                    .setCredentialsProvider(
+                            FixedCredentialsProvider.create(checkNotNull(credentials, "credentials")));
+    // We only use the batchWriteSpans API in this exporter.
+    builder
+            .batchWriteSpansSettings()
+            .setSimpleTimeoutNoRetries(org.threeten.bp.Duration.ofMillis(deadline.toMillis()));
```
nilebox, line 68: I understand that this was copied from opencensus-java, but ThreeTen is a backport of standard Duration from Java 8 to support Java 7 (https://stackoverflow.com/a/60515429).

IIRC the minimal version we intend to support is Java 8?
If that's the case, we should use https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html instead.

imnoahcook: Don't think this is possible here.
`TraceServiceSettings.setSimpleTimeoutNoRetries` is expecting a threaten duration as args.

nilebox: You're right, my bad.

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceExporter.java
imnoahcook: Sorry forgot about access modifiers for a moment. Do you think users will want to use `createWithCredentials` or `createWithClient`? I was thinking of making them private but your comment is suggesting public

nilebox: Agreed, should be sufficient to just `createWithConfiguration` public for now, otherwise it might be confusing which function to use.

PR #13, opened 2020-07-23
Add Unit Test with mock TraceServiceClient

chessegg:
# Context 

Add End to End Test ensuring that the spans generated by TraceTranslator's generateSpans() method and sent as a parameter to the batchWriteSpans() method are what we expect them to be.

# Tests
EndToEndTest.java is run with JUnit 4; can also run using `./gradlew test` on the command line

PR #16, opened 2020-07-29
Fix Rebase Issue from PR #13

chessegg:
# Context
PR 13 wasn't rebased before being merged, and some small errors popped up as a result. These are just three lines of quick fixes to get rid of those bugs so that EndToEndTest builds.


imnoahcook:
These changes are in the right direction but the test still does not build on my machine. Can someone else investigate whether it works on their machine?

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceExporter.java
```diff
@@ -71,7 +71,7 @@ private static TraceExporter createWithCredentials(
         projectId, TraceServiceClient.create(builder.build()), fixedAttributes);
   }
 
-  private TraceExporter(
+      TraceExporter(
```
imnoahcook, line 74: nit can you fix the spacing on this?

PR #17, opened 2020-07-30
Add unit tests for TraceTranslator class

chessegg:
# Context

Create unit tests to test all methods in TraceTranslator (except for generateSpans, as generateSpans is more or less just a combination of all the other methods anyways).

# Test

TraceTranslatorTest.java is run with JUnit 4; can also run using ./gradlew test on the command line

### DIFF
--- a/exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/TraceTranslatorTest.java
+++ b/exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/TraceTranslatorTest.java
@@ -1,0 +1,198 @@
+package com.google.cloud.opentelemetry.trace;
+
+import com.google.devtools.cloudtrace.v2.AttributeValue;
+import com.google.devtools.cloudtrace.v2.Span;
+import com.google.devtools.cloudtrace.v2.TruncatableString;
+import com.google.rpc.Status;
+import io.opentelemetry.common.Attributes;
+import io.opentelemetry.common.ReadableAttributes;
+import io.opentelemetry.sdk.trace.data.SpanData;
+import io.opentelemetry.trace.Span.Kind;
+import io.opentelemetry.trace.*;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+import java.util.*;
+import java.util.concurrent.TimeUnit;
+
+import static org.junit.Assert.*;
+
+@RunWith(JUnit4.class)
+public class TraceTranslatorTest {
+
+  @Test
+  public void toDisplayNameTest(){
+    String serverPrefixSpanName = "Recv. mySpanName";
+    String clientPrefixSpanName = "Sent. mySpanName";
+    String regularSpanName = "regularSpanName";
+    Kind serverSpanKind = Kind.SERVER;
+    Kind clientSpanKind = Kind.CLIENT;
+
+    assertEquals("Recv. mySpanName", TraceTranslator.toDisplayName(serverPrefixSpanName, serverSpanKind));
+    assertEquals("Sent. mySpanName", TraceTranslator.toDisplayName(clientPrefixSpanName, clientSpanKind));
+    assertEquals("Recv.regularSpanName", TraceTranslator.toDisplayName(regularSpanName, serverSpanKind));
+    assertEquals("Sent.regularSpanName", TraceTranslator.toDisplayName(regularSpanName, clientSpanKind));
+  }
+
+  @Test(expected = NullPointerException.class)
+  public void nullTruncatableStringProtoTest(){
+    TruncatableString nullTruncatable = TraceTranslator.toTruncatableStringProto(null);
+
+    assertNull(nullTruncatable.getValue());
+    assertEquals(0, nullTruncatable.getTruncatedByteCount());
+  }
+
+  @Test
+  public void toTruncatableStringProtoTest(){
+    String truncatableString = "myTruncatableString";
+    TruncatableString testTruncatable = TraceTranslator.toTruncatableStringProto(truncatableString);
+
+    assertEquals("myTruncatableString", testTruncatable.getValue());
+    assertEquals(0, testTruncatable.getTruncatedByteCount());
+  }
+
+  @Test
+  public void toTimestampProtoTest(){
+    long epochNanos = TimeUnit.SECONDS.toNanos(3001) + 255;
+    com.google.protobuf.Timestamp timestamp = TraceTranslator.toTimestampProto(epochNanos);
+
+    assertEquals(3001, timestamp.getSeconds());
+    assertEquals(255, timestamp.getNanos());
+  }
+
+  @Test
+  public void toAttributesProtoTest(){
+    ReadableAttributes attributes = Attributes.newBuilder()
+            .setAttribute("myKey", "myValue")
+            .setAttribute("http.status_code", true)
+            .setAttribute("anotherKey", 100)
+            .setAttribute("http.host", 3.14)
+            .build();
+    Map<String, AttributeValue> fixedAttributes = new LinkedHashMap<>();
+    fixedAttributes.put("fixed", AttributeValue.newBuilder().setStringValue
+        (TruncatableString.newBuilder().setValue("attributes").setTruncatedByteCount(0).build()).build());
+    fixedAttributes.put("another", AttributeValue.newBuilder().setStringValue
+        (TruncatableString.newBuilder().setValue("entry").setTruncatedByteCount(0).build()).build());
+
+    Span.Attributes.Builder attributesBuilder = Span.Attributes.newBuilder().setDroppedAttributesCount(0)
+        .putAttributeMap("myKey", AttributeValue.newBuilder().setStringValue
+            (TruncatableString.newBuilder().setValue("myValue").setTruncatedByteCount(0)).build())
+        .putAttributeMap("/http/host", AttributeValue.newBuilder().setStringValue
+            (TruncatableString.newBuilder().setValue(String.valueOf(3.14)).setTruncatedByteCount(0)).build())
+        .putAttributeMap("/http/status_code", AttributeValue.newBuilder().setBoolValue
+            (true).build())
+        .putAttributeMap("anotherKey", AttributeValue.newBuilder().setIntValue
+            (100).build())
+        .putAttributeMap("g.co/agent", AttributeValue.newBuilder().setStringValue
+            (TruncatableString.newBuilder().setValue("opentelemetry-java [0.6.0]").setTruncatedByteCount(0)).build());
+    for (Map.Entry<String, AttributeValue> entry : fixedAttributes.entrySet()) {
+      attributesBuilder.putAttributeMap(entry.getKey(), entry.getValue());
+    }
+
+    Span.Attributes allAtributes = attributesBuilder.build();
+
+    assertEquals(allAtributes, TraceTranslator.toAttributesProto(attributes, fixedAttributes));
+
+  }
+
+  @Test
+  public void toTimeEventsProtoTest(){
+    List<SpanData.Event> events = new ArrayList<>();
+    SpanData.Event eventOne = new SpanData.Event() {
+      //The SpanData.Event interfaces requires us to override these four methods
+      @Override
+      public long getEpochNanos() {
+        return 0;
+      }
+
+      @Override
+      public int getTotalAttributeCount() {
+        return 0;
+      }
+
+      @Override
+      public String getName() {
+        return "eventOne";
+      }
+
+      @Override
+      public Attributes getAttributes() {
+        return Attributes.newBuilder().setAttribute("key", "value").build();
+      }
+    };
+    events.add(eventOne);
+
+    Span.TimeEvents timeEvents = TraceTranslator.toTimeEventsProto(events);
+    assertEquals(1, timeEvents.getTimeEventCount());
+    //TODO: Figure out how to use timeEvents to call SpanData.Event methods to ensure those are working properly
+  }
+
+  @Test
+  public void toStatusProtoTest(){
+    io.opentelemetry.trace.Status myStatus = io.opentelemetry.trace.Status.OK.withDescription("Status description");
+    Status spanStatus = TraceTranslator.toStatusProto(myStatus);
+
+    //The int representation is 0 for canonical code "OK".
+    assertEquals(0, spanStatus.getCode());
+    assertEquals("Status description", spanStatus.getMessage());
+  }
+  
+  @Test
+  public void toLinksProtoTest(){
+    List<io.opentelemetry.sdk.trace.data.SpanData.Link> links = new ArrayList<>();
+
+    TraceId traceIdOne = new TraceId(321, 123);
+    SpanId spanIdOne = new SpanId(12345);
+    TraceFlags traceFlagsOne = TraceFlags.builder().build();
+    TraceState traceStateOne = TraceState.builder().build();
+    SpanContext spanContextOne = SpanContext.create(traceIdOne, spanIdOne, traceFlagsOne, traceStateOne);
+    Attributes attributesOne = Attributes.newBuilder().setAttribute("key", "value").build();
+    io.opentelemetry.sdk.trace.data.SpanData.Link linkOne =  io.opentelemetry.sdk.trace.data.SpanData.Link.create(spanContextOne, attributesOne);
+
+    TraceId traceIdTwo = new TraceId(32473, 24893);
+    SpanId spanIdTwo = new SpanId(54321);
+    TraceFlags traceFlagsTwo = TraceFlags.builder().build();
+    TraceState traceStateTwo = TraceState.builder().build();
+    SpanContext spanContextTwo = SpanContext.create(traceIdTwo, spanIdTwo, traceFlagsTwo, traceStateTwo);
+    Attributes attributesTwo = Attributes.newBuilder().setAttribute("random string", "another random string").build();
+    io.opentelemetry.sdk.trace.data.SpanData.Link linkTwo =  io.opentelemetry.sdk.trace.data.SpanData.Link.create(spanContextTwo, attributesTwo);
+
+    links.add(linkOne);
+    links.add(linkTwo);
+    Span.Links finalLinks = TraceTranslator.toLinksProto(links, 2);
+
+    assertEquals(2, finalLinks.getLinkCount());
+    assertEquals(0, finalLinks.getDroppedLinksCount());
+
+    assertEquals(traceIdOne.toLowerBase16(), finalLinks.getLink(0).getTraceId());
+    assertEquals(spanIdOne.toLowerBase16(), finalLinks.getLink(0).getSpanId());
+    //the int representation is 0 for (Link.Type.TYPE_UNSPECIFIED), which is given in TraceTranslator.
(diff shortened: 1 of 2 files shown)


exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/TraceTranslatorTest.java
```diff
@@ -0,0 +1,198 @@
...
+    assertEquals(0, nullTruncatable.getTruncatedByteCount());
+  }
+
+  @Test
+  public void toTruncatableStringProtoTest(){
+    String truncatableString = "myTruncatableString";
+    TruncatableString testTruncatable = TraceTranslator.toTruncatableStringProto(truncatableString);
+
+    assertEquals("myTruncatableString", testTruncatable.getValue());
+    assertEquals(0, testTruncatable.getTruncatedByteCount());
```
imnoahcook, line 52: when would it not be 0?

chessegg: It would be nonzero if there were bytes somehow removed from the original string

exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/TraceTranslatorTest.java
```diff
@@ -0,0 +1,198 @@
...
+import io.opentelemetry.trace.Span.Kind;
+import io.opentelemetry.trace.*;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+import java.util.*;
+import java.util.concurrent.TimeUnit;
+
+import static org.junit.Assert.*;
```
imnoahcook, line 19: prefer no star imports

exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/TraceTranslatorTest.java
```diff
@@ -0,0 +1,198 @@
...
+    //TODO: Figure out how to use timeEvents to call SpanData.Event methods to ensure those are working properly
+  }
+
+  @Test
+  public void toStatusProtoTest(){
+    io.opentelemetry.trace.Status myStatus = io.opentelemetry.trace.Status.OK.withDescription("Status description");
+    Status spanStatus = TraceTranslator.toStatusProto(myStatus);
+
+    //The int representation is 0 for canonical code "OK".
+    assertEquals(0, spanStatus.getCode());
```
imnoahcook, line 137: is there an enum you can replace 0 with for this? If not its fine

imnoahcook:
I'm not sure how I feel about turning all of the methods in `TraceTranslator` to package private. I suppose it won't affect customer's experience right?

exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/TraceTranslatorTest.java
imnoahcook: Is there a way to test that case?

chessegg: probably not with regards to `toTruncatableStringProto` actually, because the method specifies setting `truncatedByteCount` to 0.

`return TruncatableString.newBuilder().setValue(string).setTruncatedByteCount(0).build();`

exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/TraceTranslatorTest.java
chessegg: I added a test in there with the enum "OK"

exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/TraceTranslatorTest.java
chessegg: changed to specify all Trace.* and Assert.* imports

exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/TraceTranslatorTest.java
```diff
@@ -0,0 +1,198 @@
...
+    assertEquals(255, timestamp.getNanos());
+  }
+
+  @Test
+  public void toAttributesProtoTest(){
+    ReadableAttributes attributes = Attributes.newBuilder()
+            .setAttribute("myKey", "myValue")
+            .setAttribute("http.status_code", true)
+            .setAttribute("anotherKey", 100)
+            .setAttribute("http.host", 3.14)
```
skaburskis, line 76: Although totally allowed often a nice side benefit to unit tests is they can show how the code should be used. To that end I might encourage keys and values match best practices. Namely we do try to interpret http.status_code and if we get a boolean we will be confused. You could just use "boolKey"... "intKey" etc.

exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/TraceTranslatorTest.java
```diff
@@ -0,0 +1,198 @@
...
+            (true).build())
+        .putAttributeMap("anotherKey", AttributeValue.newBuilder().setIntValue
+            (100).build())
+        .putAttributeMap("g.co/agent", AttributeValue.newBuilder().setStringValue
+            (TruncatableString.newBuilder().setValue("opentelemetry-java [0.6.0]").setTruncatedByteCount(0)).build());
+    for (Map.Entry<String, AttributeValue> entry : fixedAttributes.entrySet()) {
+      attributesBuilder.putAttributeMap(entry.getKey(), entry.getValue());
+    }
+
+    Span.Attributes allAtributes = attributesBuilder.build();
```
skaburskis, line 99: typo: allAttributes

exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/TraceTranslatorTest.java
```diff
@@ -0,0 +1,198 @@
...
+
+    assertEquals(allAtributes, TraceTranslator.toAttributesProto(attributes, fixedAttributes));
```
skaburskis, line 101: I think the assertEquals will be sensitive to ordering changes and since ordering is not generally guaranteed for maps this may make this test hard to maintain or, worse, flaky. Some list matchers will allow you to specify that ordering does not matter, I don't recall if Junit4 has that option. Alternatively you can iterate over the list yourself. A slightly weaker test is preferable to a flaky one.

exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/TraceTranslatorTest.java
```diff
@@ -0,0 +1,205 @@
...
+    assertEquals(traceIdTwo.toLowerBase16(), finalLinks.getLink(1).getTraceId());
+    assertEquals(spanIdTwo.toLowerBase16(), finalLinks.getLink(1).getSpanId());
+    assertEquals(0, finalLinks.getLink(1).getTypeValue());
+    assertTrue(finalLinks.getLink(1).getAttributes().containsAttributeMap("random string"));
+    assertFalse(finalLinks.getLink(1).getAttributes().containsAttributeMap("key"));
+  }
+
+  @Test
+  public void getResourceLabelsTest(){
+    Map<String, String> nullResources = new HashMap<>();
```
nilebox, line 190: You can probably split this test into 2 separate tests - one for `nullResources` and one for non-empty set of resources. Currently you have 2 independent flows and it's hard to see how are they related to each other (they are not)

exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/TraceTranslatorTest.java
```diff
@@ -0,0 +1,205 @@
...
+
+    // The int representation is 0 for canonical code "OK".
+    assertEquals(0,myStatus.getCanonicalCode().value());
+    assertEquals(io.opentelemetry.trace.Status.CanonicalCode.OK, myStatus.getCanonicalCode());
+    assertEquals("Status description", spanStatus.getMessage());
+  }
+  
+  @Test
+  public void toLinksProtoTest(){
+    List<io.opentelemetry.sdk.trace.data.SpanData.Link> links = new ArrayList<>();
```
nilebox, line 150: `SpanData.Link` should work if you `import io.opentelemetry.sdk.trace.data.SpanData`  (here and in other places)

exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/TraceTranslatorTest.java
```diff
@@ -0,0 +1,205 @@
...
+    events.add(eventOne);
+
+    Span.TimeEvents timeEvents = TraceTranslator.toTimeEventsProto(events);
+    assertEquals(1, timeEvents.getTimeEventCount());
+    //TODO: Figure out how to use timeEvents to call SpanData.Event methods to ensure those are working properly
+  }
+
+  @Test
+  public void toStatusProtoTest(){
+    io.opentelemetry.trace.Status myStatus = io.opentelemetry.trace.Status.OK.withDescription("Status description");
```
nilebox, line 139: `import io.opentelemetry.trace.Status` unless you have a conflicting import with the same class name

exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/TraceTranslatorTest.java
```diff
@@ -0,0 +1,205 @@
...
+      @Override
+      public Attributes getAttributes() {
+        return Attributes.newBuilder().setAttribute("key", "value").build();
+      }
+    };
+    events.add(eventOne);
+
+    Span.TimeEvents timeEvents = TraceTranslator.toTimeEventsProto(events);
+    assertEquals(1, timeEvents.getTimeEventCount());
+    //TODO: Figure out how to use timeEvents to call SpanData.Event methods to ensure those are working properly
```
nilebox, line 134: Do you still want to address this?
```java
Span.TimeEvent event = timeEvents.getTimeEvent(0);
assertEquals("eventOne", event.getName());
...
```

exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/TraceTranslatorTest.java
```diff
@@ -0,0 +1,205 @@
...
+  }
+
+  @Test
+  public void toStatusProtoTest(){
+    io.opentelemetry.trace.Status myStatus = io.opentelemetry.trace.Status.OK.withDescription("Status description");
+    Status spanStatus = TraceTranslator.toStatusProto(myStatus);
+
+    // The int representation is 0 for canonical code "OK".
+    assertEquals(0,myStatus.getCanonicalCode().value());
```
haileyzhou, line 143: I believe these two lines are not really testing the "TraceTranslator .toStatusProto()" function - they're just testing whether L139 is correct, which doesn't seem to be very necessary.

I guess besides L145 "assertEquals("Status description", spanStatus.getMessage());", we might also want something like "assertEquals(0, spanStatus.getCode());"

exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/TraceTranslatorTest.java
```diff
@@ -0,0 +1,205 @@
...
+
+    TraceId traceIdOne = new TraceId(321, 123);
+    SpanId spanIdOne = new SpanId(12345);
+    TraceFlags traceFlagsOne = TraceFlags.builder().build();
+    TraceState traceStateOne = TraceState.builder().build();
+    SpanContext spanContextOne = SpanContext.create(traceIdOne, spanIdOne, traceFlagsOne, traceStateOne);
+    Attributes attributesOne = Attributes.newBuilder().setAttribute("key", "value").build();
+    io.opentelemetry.sdk.trace.data.SpanData.Link linkOne =  io.opentelemetry.sdk.trace.data.SpanData.Link.create(spanContextOne, attributesOne);
+
+    TraceId traceIdTwo = new TraceId(32473, 24893);
```
haileyzhou, line 160: This block looks pretty similar to the above one, please consider to extract out a common method to reduce verbosity.

exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/TraceTranslatorTest.java
```diff
@@ -0,0 +1,198 @@
...
+  public void toDisplayNameTest(){
+    String serverPrefixSpanName = "Recv. mySpanName";
+    String clientPrefixSpanName = "Sent. mySpanName";
+    String regularSpanName = "regularSpanName";
+    Kind serverSpanKind = Kind.SERVER;
+    Kind clientSpanKind = Kind.CLIENT;
+
+    assertEquals("Recv. mySpanName", TraceTranslator.toDisplayName(serverPrefixSpanName, serverSpanKind));
+    assertEquals("Sent. mySpanName", TraceTranslator.toDisplayName(clientPrefixSpanName, clientSpanKind));
+    assertEquals("Recv.regularSpanName", TraceTranslator.toDisplayName(regularSpanName, serverSpanKind));
```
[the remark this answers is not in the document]
chessegg: It would just return the string you sent in (namely, "Recv.regularSpanName"). The server/clientSpanKind are used to check that the string does NOT already start with "Recv." or "Sent." respectively. 
For example, if the method is passed in with parameter serverSpanKind, the method will add the literal "Recv." in front of the passed in string unless the string already starts with "Recv."

exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/TraceTranslatorTest.java
chessegg: yup, that conflicts with `import com.google.rpc.Status;`

exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/TraceTranslatorTest.java
chessegg: I tried that/similar, but the method getName() doesn't show up for me and it only displays methods of TimeEvent (and not of the SpanData.Event class). 
I also tried `TimeEvent eventOne = new SpanData.Event()` but there were issues with that too.

exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/TraceTranslatorTest.java
imnoahcook: Looks good to me then :)

exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/TraceTranslatorTest.java
chessegg: Created a method to create Links to condense the code a bit/make it more readable

exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/TraceTranslatorTest.java
chessegg: Good point, done.

exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/TraceTranslatorTest.java
```diff
@@ -0,0 +1,198 @@
...
+
+  @Test
+  public void toDisplayNameTest(){
+    String serverPrefixSpanName = "Recv. mySpanName";
+    String clientPrefixSpanName = "Sent. mySpanName";
+    String regularSpanName = "regularSpanName";
+    Kind serverSpanKind = Kind.SERVER;
+    Kind clientSpanKind = Kind.CLIENT;
+
+    assertEquals("Recv. mySpanName", TraceTranslator.toDisplayName(serverPrefixSpanName, serverSpanKind));
```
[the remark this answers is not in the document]
chessegg: Inlined the local variables in the assertEquals statement

exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/TraceTranslatorTest.java
chessegg: That's a great point. I decided to iterate over the list manually. It's a bit lengthy, but it checks that every key that I expect to be in the attributeMap is actually in there, and that its respective value is what I expect it to be. I got rid of the original chunk of code leading up to the allAttributes variable.

exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/TraceTranslatorTest.java
```diff
@@ -0,0 +1,205 @@
...
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.assertFalse;
+
+@RunWith(JUnit4.class)
+public class TraceTranslatorTest {
+
+  @Test
+  public void toDisplayNameTest(){
```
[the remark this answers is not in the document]
chessegg: changed method names to have "test" first

exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/TraceTranslatorTest.java
```diff
@@ -0,0 +1,205 @@
...
+    assertEquals("Sent. mySpanName", TraceTranslator.toDisplayName(clientPrefixSpanName, clientSpanKind));
+    assertEquals("Recv.regularSpanName", TraceTranslator.toDisplayName(regularSpanName, serverSpanKind));
+    assertEquals("Sent.regularSpanName", TraceTranslator.toDisplayName(regularSpanName, clientSpanKind));
+  }
+
+  @Test(expected = NullPointerException.class)
+  public void nullTruncatableStringProtoTest(){
+    TruncatableString nullTruncatable = TraceTranslator.toTruncatableStringProto(null);
+
+    assertNull(nullTruncatable.getValue());
```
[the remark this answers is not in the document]
chessegg: Got it. It seems that for this test, just testing that the NullPointerException is thrown is enough, so I added the line of code you suggested at the bottom.

exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/TraceTranslatorTest.java
chessegg: Makes sense, changed

exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/TraceTranslatorTest.java
chessegg: Added variables (stringKey, boolKey, longKey, doubleKey) and then put those variables into the parameters of `setAttribute` instead

nilebox:
Looks good enough to be merged, can be improved further with separate PRs.

exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/TraceTranslatorTest.java
nilebox: Fixed in #18

PR #19, opened 2020-07-30
Switch EndToEndTest to use mock server

the-ericwang35:
We'd like to integrate our [mock server](https://github.com/googleinterns/cloud-operations-api-mock) into the Java exporter to be used for integration testing.

Currently, the integration test just calls the mock server with the translated spans to verify that they follow the correct format and that the exporter has translated the spans properly. In the future, more tests can be added as needed.

### DIFF
--- a/exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/MockCloudTraceClient.java
+++ b/exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/MockCloudTraceClient.java
@@ -10,15 +10,15 @@ import io.grpc.ManagedChannelBuilder;
 import java.util.List;
 
 // A simplified version of TraceServiceClient, used ONLY for testing purposes.
-public class MockTraceServiceClient {
+public class MockCloudTraceClient implements CloudTraceClient{
 
     TraceServiceGrpc.TraceServiceBlockingStub blockingStub;
 
-    public MockTraceServiceClient(String host, int port) {
+    public MockCloudTraceClient(String host, int port) {
         this(ManagedChannelBuilder.forAddress(host, port).usePlaintext());
     }
 
-    public MockTraceServiceClient(ManagedChannelBuilder<?> channelBuilder) {
+    public MockCloudTraceClient(ManagedChannelBuilder<?> channelBuilder) {
         Channel channel = channelBuilder.build();
         blockingStub = TraceServiceGrpc.newBlockingStub(channel);
     }
--- a/exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/MockTraceServiceClient.java
+++ b/exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/MockTraceServiceClient.java
@@ -9,6 +9,7 @@ import io.grpc.ManagedChannelBuilder;
 
 import java.util.List;
 
+// A simplified version of TraceServiceClient, used ONLY for testing purposes.
 public class MockTraceServiceClient {
 
     TraceServiceGrpc.TraceServiceBlockingStub blockingStub;
--- a/exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/MockTraceServiceClient.java
+++ b/exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/MockTraceServiceClient.java
@@ -1,0 +1,33 @@
+package com.google.cloud.opentelemetry.trace;
+
+import com.google.devtools.cloudtrace.v2.BatchWriteSpansRequest;
+import com.google.devtools.cloudtrace.v2.ProjectName;
+import com.google.devtools.cloudtrace.v2.Span;
+import com.google.devtools.cloudtrace.v2.TraceServiceGrpc;
+import io.grpc.Channel;
+import io.grpc.ManagedChannelBuilder;
+
+import java.util.List;
+
+public class MockTraceServiceClient {
+
+    TraceServiceGrpc.TraceServiceBlockingStub blockingStub;
+
+    public MockTraceServiceClient(String host, int port) {
+        this(ManagedChannelBuilder.forAddress(host, port).usePlaintext());
+    }
+
+    public MockTraceServiceClient(ManagedChannelBuilder<?> channelBuilder) {
+        Channel channel = channelBuilder.build();
+        blockingStub = TraceServiceGrpc.newBlockingStub(channel);
+    }
+
+    public final void batchWriteSpans(ProjectName name, List<Span> spans) {
+        BatchWriteSpansRequest request =
+                BatchWriteSpansRequest.newBuilder()
+                        .setName(name.toString())
+                        .addAllSpans(spans)
+                        .build();
+        blockingStub.batchWriteSpans(request);
+    }
+}
--- a/exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceExporter.java
+++ b/exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceExporter.java
@@ -23,8 +23,7 @@ import static com.google.api.client.util.Preconditions.checkNotNull;
 
 public class TraceExporter implements SpanExporter {
 
-  private final TraceServiceClient traceServiceClient;
-  private final MockTraceServiceClient mockTraceServiceClient;
+  private final CloudTraceClient cloudTraceClient;
   private final ProjectName projectName;
   private final String projectId;
   private final Map<String, AttributeValue> fixedAttributes;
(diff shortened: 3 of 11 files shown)


exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/MockTraceServiceClient.java
```diff
@@ -0,0 +1,33 @@
...
+import com.google.devtools.cloudtrace.v2.BatchWriteSpansRequest;
+import com.google.devtools.cloudtrace.v2.ProjectName;
+import com.google.devtools.cloudtrace.v2.Span;
+import com.google.devtools.cloudtrace.v2.TraceServiceGrpc;
+import io.grpc.Channel;
+import io.grpc.ManagedChannelBuilder;
+
+import java.util.List;
+
+public class MockTraceServiceClient {
```
the-ericwang35, line 12: Unlike Python, `TraceServiceClient` was very difficult to work with, as it has a lot of setup involving TLS/SSL, GCP credentials, etc. and I was unable to get it to connect to the mock server. So, I created this simplified version of it to be used for tests

exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/EndToEndTest.java
```diff
@@ -89,33 +95,8 @@ public void exportMockSpanDataList(){
...
-    // Instead of doReturn, doAnswer is better to deal with whenever batchWriteSpans() is called.
-    // Rather than actually call the BatchWriteSpans() method from the API, null is returned.
-    doAnswer(invocation -> null)
-            .when(mockTraceServiceClient).batchWriteSpans(eq(PROJECT_ID), anyList());
-
     // Invokes export();
     assertEquals(SpanExporter.ResultCode.SUCCESS, exporter.export(spanDataList));
-
-    // Make sure that the parameters passed to batchWriteSpans() are what we expect them to be
-    verify(mockTraceServiceClient, times(1)).batchWriteSpans(eq(ProjectName.of(PROJECT_ID)),
```
the-ericwang35, line 117: I removed this check since `verify` is a `Mockito` thing. This shouldn't be needed anymore since the export won't be successful if the span translation didn't work properly (the mock server has validations for these), and there are unit tests for `TraceTranslator` now

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceTranslator.java
```diff
@@ -28,9 +28,11 @@
 
   // TODO(nilebox): Extract the constant
   private static final String OPEN_TELEMETRY_LIBRARY_VERSION = "0.6.0";
+  private static final String EXPORTER_VERSION = "0.1.0";
   private static final String AGENT_LABEL_KEY = "g.co/agent";
   private static final String AGENT_LABEL_VALUE_STRING =
-      "opentelemetry-java [" + OPEN_TELEMETRY_LIBRARY_VERSION + "]";
+      "opentelemetry-java " + OPEN_TELEMETRY_LIBRARY_VERSION +
```
the-ericwang35, line 34: The format of the agent attribute has been changed to `opentelemetry-<language_code> <ot_version>; google-cloud-trace-exporter <exporter_version>` (see [here](https://docs.google.com/document/d/1eeGeSYlE2F-Dc-5_yLXnIlj7e--BYFeWlzWy6qBV2cg/edit#heading=h.aoqug1blgfpl))

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/MockTraceServiceClient.java
nilebox: That's really unfortunate, would be nice to find a way to use the standard client with mock server.
Fine to work on that after this PR is merged though.

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceExporter.java
```diff
@@ -24,6 +24,7 @@
 public class TraceExporter implements SpanExporter {
 
   private final TraceServiceClient traceServiceClient;
+  private final MockTraceServiceClient mockTraceServiceClient;
```
nilebox, line 27: `TraceExporter` shouldn't know anything about mocking.
If you want to be able to replace `TraceServiceClient` with `MockTraceServiceClient`, you need to define an interface abstraction and implement it for mock and for production (wrapper around TraceServiceClient).

e.g.
```java
public interface CloudTraceClient {
  void batchWriteSpans(ProjectName name, List<Span> spans)
}

public class MockCloudTraceClient implements CloudTraceClient {
  ...
}

public class CloudTraceClientImpl implements CloudTraceClient {
  private final TraceServiceClient traceServiceClient;
  ...

  public final void batchWriteSpans(ProjectName name, List<Span> spans) {
    traceServiceClient.batchWriteSpans(name, spans);
  }
}
```

Does it make sense?

exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/EndToEndTest.java
nilebox: Could you also delete https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/tree/master/exporters/trace/src/test/resources/mockito-extensions please?

nilebox: And also delete dependency on mockito in https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/blob/6f0beb2bed141677fc37911fa4855120ec189673/exporters/trace/build.gradle#L11 and https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/blob/6f0beb2bed141677fc37911fa4855120ec189673/build.gradle#L34

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceExporter.java
the-ericwang35: Yep good idea, I've made the change. Please let me know if this was not what you were describing.

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/MockCloudTraceClient.java
```diff
@@ -0,0 +1,34 @@
...
+import com.google.devtools.cloudtrace.v2.ProjectName;
+import com.google.devtools.cloudtrace.v2.Span;
+import com.google.devtools.cloudtrace.v2.TraceServiceGrpc;
+import io.grpc.Channel;
+import io.grpc.ManagedChannelBuilder;
+
+import java.util.List;
+
+// A simplified version of TraceServiceClient, used ONLY for testing purposes.
+public class MockCloudTraceClient implements CloudTraceClient{
```
nilebox, line 13: This class could be package private I think

exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/EndToEndTest.java
```diff
@@ -38,38 +36,46 @@
...
-  public void setup() {
-    MockitoAnnotations.initMocks(this);
+  public void setup() throws IOException {
+    // Find a free port to spin up our server at.
+    ServerSocket socket = new ServerSocket(0);
+    int port = socket.getLocalPort();
+    String address = String.format("%s:%d", LOCALHOST, port);
+    socket.close();
+
+    // Start the mock server. This assumes the binary is present and in $PATH.
```
nilebox, line 56: So once we create the CircleCI build we will need to install the binary before executing this test, correct?

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/MockCloudTraceClient.java
the-ericwang35: Done

exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/EndToEndTest.java
the-ericwang35: Yes, what the Python exporter does is have a shell script `curl` the release binary from our repo, and then add it to `$PATH` (Python uses `Tox` to add the binary to `{envbindir}`, which is already part of `$PATH`, CircleCI probably needs to manually do this). The script then creates a symlink for the binary to avoid having to deal with versions. See [this PR](https://github.com/GoogleCloudPlatform/opentelemetry-operations-python/pull/30) for the script).

PR #20, opened 2020-07-31
Add an example project

imnoahcook:
# Context
Right now customer don't have any way to see how to use our exporter

# Solution 
Create a TraceExample project with some sample code that customers can use. This involves a new gradle project with `build.gradle` and dependencies

# Test plan
Automatic tests
```sh
$ gradle test && gradle build
```
In addition I tested this project by deploying it to GKE using the following dockerfile
```dockerfile
FROM openjdk:8
VOLUME /tmp
RUN mkdir /application
COPY . /application
WORKDIR /application
RUN /application/gradlew build
RUN /application/gradlew clean build examples-trace:fatJar #if you use maven you can change to mvnw
RUN mv examples/trace/build/libs/examples-trace-0.0.0-SNAPSHOT-all.jar /application/app.jar
ENTRYPOINT ["java","-Djava.security.egd=file:/dev/./urandom","-jar","/application/app.jar"]
```
And list of commands
```sh
docker build --no-cache -t traceexample . &&
gcloud container clusters create traceexample-cluster --num-nodes=3 --region=us-central1-a &&
gcloud builds submit --tag=gcr.io/noahcook/traceexample:v1 . &&
gcloud config set container/cluster traceexample-cluster &&
kubectl run traceexample --image=gcr.io/noahcook/traceexample:v1 --port 8080 &&
kubectl get pods &&
kubectl expose deployment traceexample --type=LoadBalancer --port 80 --target-port 8080 &&
kubectl get service
```
The stuff about the ports is because I used this deploy script for a different project.
After being deployed the traces show up on gcloud.

# Future Work
I'll also need to add a ReadMe for this, but I'll wait until this is merged

### DIFF
--- a/examples/trace/src/main/java/com/google/cloud/opentelemetry/example/trace/TraceExample.java
+++ b/examples/trace/src/main/java/com/google/cloud/opentelemetry/example/trace/TraceExample.java
@@ -1,0 +1,51 @@
+package com.google.cloud.opentelemetry.example.trace;
+
+import com.google.cloud.opentelemetry.trace.TraceConfiguration;
+import com.google.cloud.opentelemetry.trace.TraceExporter;
+import com.google.cloud.trace.v2.TraceServiceClient;
+import io.opentelemetry.OpenTelemetry;
+import io.opentelemetry.sdk.OpenTelemetrySdk;
+//import io.opentelemetry.sdk.trace.export.SimpleSpansProcessor;
+import io.opentelemetry.trace.Span;
+import io.opentelemetry.trace.Tracer;
+
+import java.io.IOException;
+
+public class TraceExample {
+  private TraceExporter traceExporter;
+
+  // OTel API
+  private Tracer tracer = OpenTelemetry.getTracer("io.opentelemetry.example.TraceExample");
+
+  private void setupTraceExporter() {
+    
+    // using default configuration
+    TraceConfiguration configuration = TraceConfiguration.builder().build();
+
+//    OpenTelemetrySdk.getTracerProvider()
+//            .addSpanProcessor(SimpleSpansProcessor.newBuilder(this.traceExporter).build());
+  }
+
+  private void myUseCase() {
+    // Generate a span
+    Span span = this.tracer.spanBuilder("Start my wonderful use case").startSpan();
+    span.addEvent("Event 0");
+    // execute my use case - here we simulate a wait
+    doWork();
+    span.addEvent("Event 1");
+    span.end();
+  }
+
+  private void doWork() {
+    try {
+      Thread.sleep((int) (Math.random() * 1000) + 1000);
+    } catch (InterruptedException e) {
+    }
+  }
+
+  public static void main(String[] args) {
+    TraceExample example = new TraceExample();
+    example.setupTraceExporter();
+    // System.out.println("TODO: create some spans");
+  }
+}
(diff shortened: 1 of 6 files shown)


examples/trace/src/main/java/com/google/cloud/opentelemetry/example/trace/TraceExample.java
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  public static void main(String[] args) {
+    TraceExample example = new TraceExample();
+    example.setupTraceExporter();
+
+    while (true) {
+      example.myUseCase();
+      try {
+        // one hour
```
haileyzhou, line 58: nit: please be explicit why sleep for an hour here. Since this class will be used as a public example for users, we may need to include more explanations to help users understand.

examples/trace/src/main/java/com/google/cloud/opentelemetry/example/trace/TraceExample.java
```diff
@@ -0,0 +1,64 @@
...
+        Thread.sleep(1000 * 60 * 60);
+      } catch (InterruptedException e) {
```
haileyzhou, line 60: try {
...
} catch (InterruptedException e) {
} 

so we dont lose the exception.

examples/trace/src/main/java/com/google/cloud/opentelemetry/example/trace/TraceExample.java
```diff
@@ -0,0 +1,64 @@
...
+    // generate a span
+    Span span = this.tracer.spanBuilder("Start my use case").startSpan();
+    span.addEvent("Event 0");
+    // simulate work
+    doWork();
+    span.addEvent("Event 1");
+    span.end();
+  }
+
+  private void doWork() {
```
haileyzhou, line 44: May I ask what's the goal of doWork()?

examples/trace/src/main/java/com/google/cloud/opentelemetry/example/trace/TraceExample.java
```diff
@@ -0,0 +1,64 @@
...
+import io.opentelemetry.OpenTelemetry;
+import io.opentelemetry.sdk.OpenTelemetrySdk;
+import io.opentelemetry.sdk.trace.export.SimpleSpanProcessor;
+import io.opentelemetry.trace.Span;
+import io.opentelemetry.trace.Tracer;
+
+import java.io.IOException;
+import java.time.Duration;
+
+public class TraceExample {
```
haileyzhou, line 14: TraceExample sounds a bit too general for me, how about TraceExporterExample?

examples/trace/src/main/java/com/google/cloud/opentelemetry/example/trace/TraceExample.java
imnoahcook: `doWork()` will sleep for a random amount of time to simulate some sort of work, so not every span being exported has the same length, and in addition the span exists for a nonzero amount of time.

examples/trace/src/main/java/com/google/cloud/opentelemetry/example/trace/TraceExample.java
imnoahcook: Good suggestion

examples/trace/src/main/java/com/google/cloud/opentelemetry/example/trace/TraceExample.java
imnoahcook: I'm having it run for hours to show more traces. perhaps it would make more sense to just have one span and export it instead of going in a loop?

imnoahcook: I ended up just exporting the single span

PR #26, opened 2020-08-14
Add CircleCI build and test

zoercai:
This adds CircleCI which builds and runs tests.


zoercai:
@nilebox @james-bebbington

PR #28, opened 2020-08-18
Start 0.8.0 development cycle

zoercai:
Bump master to the next version. 

Please don't merge this until the 0.7.0 deployment is released to the Central Repository.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -5,7 +5,7 @@ subprojects {
     apply plugin: 'signing'
 
     group = "com.google.cloud.opentelemetry"
-    version = "0.0.0-SNAPSHOT" // CURRENT_VERSION
+    version = "0.8.0-SNAPSHOT" // CURRENT_VERSION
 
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -5,7 +5,7 @@ subprojects {
     apply plugin: 'signing'
 
     group = "com.google.cloud.opentelemetry"
-    version = "0.0.0-SNAPSHOT" // CURRENT_VERSION
+    version = "0.8.0-SNAPSHOT" // CURRENT_VERSION
```
nilebox, line 8: Shouldn't it be `v0.6.0` based on the current dependency version https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/blob/ffcdeb99466b848ae1138b68ce431187e0ac7039/build.gradle#L22 ?

zoercai: I've updated the `openTelemetryVersion` to 0.7.0 on master, so 0.8.0 should be correct now

nilebox (approved):

PR #31, opened 2020-08-21
Add release instruction

zoercai:


RELEASING.md
```diff
@@ -0,0 +1,194 @@
...
+    key](http://central.sonatype.org/pages/working-with-pgp-signatures.html#distributing-your-public-key)
+    to make it visible to the Sonatype servers.
+- Put your GnuPG key password and OSSRH account information in
+  `<your-home-directory>/.gradle/gradle.properties`:
+
+    ```text
+    # You need the signing properties only if you are making release deployment
+    signing.keyId=<8-character-public-key-id>
+    signing.password=<key-password>
+    signing.secretKeyRingFile=<your-home-directory>/.gnupg/secring.gpg
```
nilebox, line 38: Shall we add a caveat about extracting secring from newer versions of GPG? Or just add a link to https://docs.gradle.org/current/userguide/signing_plugin.html#sec:signatory_credentials for up-to-date instructions.

RELEASING.md
```diff
@@ -0,0 +1,194 @@
...
+    checkstyle.ignoreFailures=false
+    ```
+
+## Download the mock server
+
+- Download the [mock server
+  executable](https://github.com/googleinterns/cloud-operations-api-mock/releases),
+  and save the path.
+
+    ```bash
```
nilebox, line 51: Since you've extracted this into https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/blob/master/get_mock_server.sh, I think we may replace the instructions with just 
```bash
./get_mock_server.sh
```

RELEASING.md
```diff
@@ -0,0 +1,194 @@
...
+    $ VERSION=v2-alpha
+    $ BINARY=mock_server-x64-linux-$VERSION
+    $ curl -L -o $BINARY https://github.com/googleinterns/cloud-operations-api-mock/releases/download/$VERSION/$BINARY
+    $ chmod +x $BINARY
+    $ MOCKSERVER=$PWD/$BINARY
```
nilebox, line 56: re: suggestion to invoke a script above. You're using this env var below, so we need to keep it.
or we could add this line to the bash script and invoke it as `source ./get_mock_server.sh`

PR #32, opened 2020-08-21
Use released dependency for example project

zoercai:


build.gradle
```diff
@@ -21,15 +21,17 @@ subprojects {
...
-                google_cloud_trace    : "com.google.cloud:google-cloud-trace:${googleCloudVersion}",
-                opentelemetry_api     : "io.opentelemetry:opentelemetry-api:${openTelemetryVersion}",
-                opentelemetry_sdk     : "io.opentelemetry:opentelemetry-sdk:${openTelemetryVersion}",
+                auto_value_annotations       : "com.google.auto.value:auto-value-annotations:${autoValueVersion}",
+                auto_value                   : "com.google.auto.value:auto-value:${autoValueVersion}",
+                google_cloud_core            : "com.google.cloud:google-cloud-core:${googleCloudVersion}",
+                google_cloud_grpc            : "com.google.api.grpc:grpc-google-cloud-trace-v2:${googleCloudVersion}",
+                google_cloud_trace           : "com.google.cloud:google-cloud-trace:${googleCloudVersion}",
+                opentelemetry_api            : "io.opentelemetry:opentelemetry-api:${openTelemetryVersion}",
+                opentelemetry_sdk            : "io.opentelemetry:opentelemetry-sdk:${openTelemetryVersion}",
```
zoercai, line 33: just indentation changes here

build.gradle
```diff
@@ -65,6 +67,9 @@ subprojects {
     }
 
     uploadArchives {
+        onlyIf {
+            name in ['exporter-trace']
+        }
```
zoercai, line 72: This is to exclude the example project from being released

RELEASING.md
```diff
@@ -79,6 +79,8 @@ token](https://help.github.com/articles/creating-a-personal-access-token-for-the
     $ git checkout -b bump-version master
     # Change version to next minor (and keep -SNAPSHOT)
     $ sed -i 's/[0-9]\+\.[0-9]\+\.[0-9]\+\(.*CURRENT_VERSION\)/'$MAJOR.$((MINOR+1)).0'\1/' build.gradle
+    # Update the example project dependency to this release version
+   $ sed -i ’s/[0-9]\+\.[0-9]\+\.[0-9]\+\(.*CURRENT_RELEASE_VERSION\)/‘$MAJOR.$MINOR.$PATCH’\1/‘ build.gradle
```
nilebox: Nice 👍

examples/trace/build.gradle
```diff
@@ -24,12 +23,11 @@ task fatJar(type: Jar) {
...
 }
 
-
 description = 'Examples for Cloud Trace Exporter'
 
 dependencies {
     compile(libraries.opentelemetry_api)
     compile(libraries.opentelemetry_sdk)
     compile(libraries.google_cloud_trace)
-    compile project(':exporter-trace')
```
nilebox, line 34: 👍

nilebox: One thing I'm not too sure about here is how will we deal with breaking changes. Will we have to update example only after exporter is released? That's probably fine, just untrivial to test before making a release.

zoercai: Yeah, the changes to the example project code for breaking changes will be on the `bump-version` branch, which is also where the version update is for the example project dependency, so it shouldn't block the release process. I think it should be okay to carry on with the release, and come back to the `bump-version` branch once the release is complete to fix up the example project code. WDYT?
I could add a comment in the `bump-version` section of the release instructions to say the above.

nilebox: Sounds good to me!

PR #33, opened 2020-08-31
Add metrics exporter

zoercai:
This adds the metric exporter for Google cloud monitoring.
Unit tests here: https://github.com/zoercai/opentelemetry-operations-java/pull/1
End to end tests (WIP) here: https://github.com/zoercai/opentelemetry-operations-java/pull/2

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -19,6 +19,7 @@ subprojects {
     ext {
         autoValueVersion = '1.7.4'
         googleCloudVersion = '1.0.2'
+        cloudMonitoringVersion = '2.0.1'
         openTelemetryVersion = '0.7.0'
         junitVersion = '4.13'
         opentelemetryOperationsVersion = "0.7.1"  // CURRENT_RELEASE_VERSION
@@ -29,6 +30,7 @@ subprojects {
                 google_cloud_core            : "com.google.cloud:google-cloud-core:${googleCloudVersion}",
                 google_cloud_grpc            : "com.google.api.grpc:grpc-google-cloud-trace-v2:${googleCloudVersion}",
                 google_cloud_trace           : "com.google.cloud:google-cloud-trace:${googleCloudVersion}",
+                google_cloud_monitoring      : "com.google.cloud:google-cloud-monitoring:${cloudMonitoringVersion}",
                 opentelemetry_api            : "io.opentelemetry:opentelemetry-api:${openTelemetryVersion}",
                 opentelemetry_sdk            : "io.opentelemetry:opentelemetry-sdk:${openTelemetryVersion}",
                 opentelemetry_operations_java: "com.google.cloud.opentelemetry:exporter-trace:${opentelemetryOperationsVersion}"
--- a/build.gradle
+++ b/build.gradle
@@ -18,6 +18,7 @@ subprojects {
 
     ext {
         autoValueVersion = '1.7.4'
+        slf4jVersion = '1.7.30'
         googleCloudVersion = '1.0.2'
         cloudMonitoringVersion = '2.0.1'
         openTelemetryVersion = '0.7.0'
@@ -28,9 +29,11 @@ subprojects {
                 auto_value_annotations       : "com.google.auto.value:auto-value-annotations:${autoValueVersion}",
                 auto_value                   : "com.google.auto.value:auto-value:${autoValueVersion}",
                 google_cloud_core            : "com.google.cloud:google-cloud-core:${googleCloudVersion}",
-                google_cloud_grpc            : "com.google.api.grpc:grpc-google-cloud-trace-v2:${googleCloudVersion}",
                 google_cloud_trace           : "com.google.cloud:google-cloud-trace:${googleCloudVersion}",
+                google_cloud_trace_grpc      : "com.google.api.grpc:grpc-google-cloud-trace-v2:${googleCloudVersion}",
                 google_cloud_monitoring      : "com.google.cloud:google-cloud-monitoring:${cloudMonitoringVersion}",
+                google_cloud_monitoring_grpc : "com.google.cloud:grpc-google-cloud-monitoring-v3:${cloudMonitoringVersion}",
+                slf4j                        : "org.slf4j:slf4j-api:${slf4jVersion}",
                 opentelemetry_api            : "io.opentelemetry:opentelemetry-api:${openTelemetryVersion}",
                 opentelemetry_sdk            : "io.opentelemetry:opentelemetry-sdk:${openTelemetryVersion}",
                 opentelemetry_operations_java: "com.google.cloud.opentelemetry:exporter-trace:${opentelemetryOperationsVersion}"
--- a/build.gradle
+++ b/build.gradle
@@ -39,7 +39,9 @@ subprojects {
                 opentelemetry_operations_java: "com.google.cloud.opentelemetry:exporter-trace:${opentelemetryOperationsVersion}"
         ]
         testLibraries = [
-                junit: "junit:junit:${junitVersion}",
+                junit       : "junit:junit:${junitVersion}",
+                mockito     : "org.mockito:mockito-inline:${mockitoVersion}",
+                slf4j_simple: "org.slf4j:slf4j-simple:${slf4jVersion}"
         ]
     }
 
--- a/build.gradle
+++ b/build.gradle
@@ -23,6 +23,7 @@ subprojects {
         cloudMonitoringVersion = '2.0.1'
         openTelemetryVersion = '0.7.0'
         junitVersion = '4.13'
+        mockitoVersion = '3.5.10'
         opentelemetryOperationsVersion = "0.7.1"  // CURRENT_RELEASE_VERSION
 
         libraries = [
--- a/build.gradle
+++ b/build.gradle
@@ -41,7 +41,6 @@ subprojects {
         ]
         testLibraries = [
                 junit       : "junit:junit:${junitVersion}",
-                mockito     : "org.mockito:mockito-inline:${mockitoVersion}",
                 slf4j_simple: "org.slf4j:slf4j-simple:${slf4jVersion}"
         ]
     }
--- a/build.gradle
+++ b/build.gradle
@@ -23,7 +23,6 @@ subprojects {
         cloudMonitoringVersion = '2.0.1'
         openTelemetryVersion = '0.7.0'
         junitVersion = '4.13'
-        mockitoVersion = '3.5.10'
         opentelemetryOperationsVersion = "0.7.1"  // CURRENT_RELEASE_VERSION
 
         libraries = [
--- a/exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricConfiguration.java
+++ b/exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricConfiguration.java
@@ -1,24 +1,31 @@
 package com.google.cloud.opentelemetry.metric;
 
+import static java.time.Duration.ZERO;
+
 import com.google.auth.Credentials;
 import com.google.auto.value.AutoValue;
 import com.google.cloud.ServiceOptions;
 import com.google.cloud.monitoring.v3.stub.MetricServiceStub;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
+import java.time.Duration;
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.Immutable;
 
-/** Configurations for {@link MetricExporter}. */
+/**
+ * Configurations for {@link MetricExporter}.
+ */
 @AutoValue
 @Immutable
 public abstract class MetricConfiguration {
 
     private static final String DEFAULT_PROJECT_ID =
-            Strings.nullToEmpty(ServiceOptions.getDefaultProjectId());
+        Strings.nullToEmpty(ServiceOptions.getDefaultProjectId());
     private static final boolean DEFAULT_ADD_UNIQUE_IDENTIFIER = false;
+    private static final Duration DEFAULT_DEADLINE = Duration.ofSeconds(10, 0);
 
-    MetricConfiguration() {}
+    MetricConfiguration() {
+    }
 
     /**
      * Returns the {@link Credentials}.
@@ -50,21 +57,36 @@ public abstract class MetricConfiguration {
     @Nullable
     public abstract MetricServiceStub getMetricServiceStub();
 
+    /**
+     * Returns the deadline for exporting to Cloud Monitoring backend.
+     *
+     * <p>Default value is 10 seconds.
+     *
+     * @return the export deadline.
+     */
+    public abstract Duration getDeadline();
+
     public static Builder builder() {
         return new AutoValue_MetricConfiguration.Builder()
             .setProjectId(DEFAULT_PROJECT_ID)
-            .setAddUniqueIdentifier(DEFAULT_ADD_UNIQUE_IDENTIFIER);
+            .setAddUniqueIdentifier(DEFAULT_ADD_UNIQUE_IDENTIFIER)
+            .setDeadline(DEFAULT_DEADLINE);
     }
 
-    /** Builder for {@link MetricConfiguration}. */
+    /**
+     * Builder for {@link MetricConfiguration}.
+     */
     @AutoValue.Builder
     public abstract static class Builder {
 
-        Builder() {}
+        Builder() {
+        }
 
         abstract String getProjectId();
 
         abstract boolean getAddUniqueIdentifier();
+
+        abstract Duration getDeadline();
 
         public abstract Builder setProjectId(String projectId);
 
@@ -73,6 +95,8 @@ public abstract class MetricConfiguration {
         public abstract Builder setAddUniqueIdentifier(boolean newAddUniqueIdentifier);
 
         public abstract Builder setMetricServiceStub(MetricServiceStub newMetricServiceStub);
+
+        public abstract Builder setDeadline(Duration deadline);
 
         abstract MetricConfiguration autoBuild();
 
@@ -83,9 +107,9 @@ public abstract class MetricConfiguration {
          */
         public MetricConfiguration build() {
             Preconditions.checkArgument(
-                    !Strings.isNullOrEmpty(getProjectId()),
-                    "Cannot find a project ID from either configurations or application default.");
-
+                !Strings.isNullOrEmpty(getProjectId()),
+                "Cannot find a project ID from either configurations or application default.");
+            Preconditions.checkArgument(getDeadline().compareTo(ZERO) > 0, "Deadline must be positive.");
             return autoBuild();
         }
     }
--- a/exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricConfiguration.java
+++ b/exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricConfiguration.java
@@ -1,0 +1,98 @@
+package com.google.cloud.opentelemetry.metric;
+
+import com.google.auth.Credentials;
+import com.google.auto.value.AutoValue;
+import com.google.cloud.ServiceOptions;
+import com.google.cloud.monitoring.v3.MetricServiceClient;
+import com.google.cloud.monitoring.v3.stub.MetricServiceStub;
+import com.google.common.base.Preconditions;
+import com.google.common.base.Strings;
+
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.Immutable;
+
+/** Configurations for {@link MetricExporter}. */
+@AutoValue
+@Immutable
+public abstract class MetricConfiguration {
+
+    private static final String DEFAULT_PROJECT_ID =
+            Strings.nullToEmpty(ServiceOptions.getDefaultProjectId());
+    private static final boolean DEFAULT_ADD_UNIQUE_IDENTIFIER = false;
+
+    MetricConfiguration() {}
+
+    /**
+     * Returns the {@link Credentials}.
+     *
+     * @return the {@code Credentials}.
+     */
+    @Nullable
+    public abstract Credentials getCredentials();
+
+    /**
+     * Returns the cloud project id.
+     *
+     * @return the cloud project id.
+     */
+    public abstract String getProjectId();
+
+    /**
+     * Returns the client.
+     *
+     * @return the client.
+     */
+    public abstract MetricServiceClient getClient();
+
+    /**
+     * Returns whether a unique identifier will be added.
+     *
+     * @return whether a unique identifier will be added.
+     */
+    public abstract boolean getAddUniqueIdentifier();
+
+    /**
+     * Returns a MetricsServiceStub instance used to make RPC calls.
+     *
+     * @return the metrics service stub.
+     */
+    @Nullable
+    public abstract MetricServiceStub getMetricServiceStub();
+
+    public static Builder builder() {
+        return new AutoValue_MetricConfiguration.Builder();
+    }
+
+    /** Builder for {@link MetricConfiguration}. */
+    @AutoValue.Builder
+    public abstract static class Builder {
+
+        Builder() {}
+
+        abstract String getProjectId();
+
+        abstract MetricServiceClient getClient();
+
+        abstract boolean getAddUniqueIdentifier();
+
+        abstract MetricConfiguration autoBuild();
+
+        public abstract Builder setProjectId(String newProjectId);
+
+        public abstract Builder setClient(MetricServiceClient newClient);
+
+        /**
+         * Builds a {@link MetricConfiguration}.
+         *
+         * @return a {@code MetricsConfiguration}.
+         */
+        public MetricConfiguration build() {
+            Preconditions.checkArgument(
+                    !Strings.isNullOrEmpty(getProjectId()),
+                    "Cannot find a project ID from either configurations or application default.");
+
+            return autoBuild();
+        }
+    }
+
+}
--- a/exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricConfiguration.java
+++ b/exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricConfiguration.java
@@ -3,11 +3,9 @@ package com.google.cloud.opentelemetry.metric;
 import com.google.auth.Credentials;
 import com.google.auto.value.AutoValue;
 import com.google.cloud.ServiceOptions;
-import com.google.cloud.monitoring.v3.MetricServiceClient;
 import com.google.cloud.monitoring.v3.stub.MetricServiceStub;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
-
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.Immutable;
 
@@ -38,13 +36,6 @@ public abstract class MetricConfiguration {
     public abstract String getProjectId();
 
     /**
-     * Returns the client.
-     *
-     * @return the client.
-     */
-    public abstract MetricServiceClient getClient();
-
-    /**
      * Returns whether a unique identifier will be added.
      *
      * @return whether a unique identifier will be added.
@@ -60,7 +51,9 @@ public abstract class MetricConfiguration {
     public abstract MetricServiceStub getMetricServiceStub();
 
     public static Builder builder() {
-        return new AutoValue_MetricConfiguration.Builder();
+        return new AutoValue_MetricConfiguration.Builder()
+            .setProjectId(DEFAULT_PROJECT_ID)
+            .setAddUniqueIdentifier(DEFAULT_ADD_UNIQUE_IDENTIFIER);
     }
 
     /** Builder for {@link MetricConfiguration}. */
@@ -71,15 +64,17 @@ public abstract class MetricConfiguration {
 
         abstract String getProjectId();
 
-        abstract MetricServiceClient getClient();
-
         abstract boolean getAddUniqueIdentifier();
 
+        public abstract Builder setProjectId(String projectId);
+
+        public abstract Builder setCredentials(Credentials newCredentials);
+
+        public abstract Builder setAddUniqueIdentifier(boolean newAddUniqueIdentifier);
+
+        public abstract Builder setMetricServiceStub(MetricServiceStub newMetricServiceStub);
+
         abstract MetricConfiguration autoBuild();
-
-        public abstract Builder setProjectId(String newProjectId);
-
-        public abstract Builder setClient(MetricServiceClient newClient);
 
         /**
          * Builds a {@link MetricConfiguration}.
--- a/exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricConfiguration.java
+++ b/exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricConfiguration.java
@@ -21,7 +21,6 @@ public abstract class MetricConfiguration {
 
     private static final String DEFAULT_PROJECT_ID =
         Strings.nullToEmpty(ServiceOptions.getDefaultProjectId());
-    private static final boolean DEFAULT_ADD_UNIQUE_IDENTIFIER = false;
     private static final Duration DEFAULT_DEADLINE = Duration.ofSeconds(10, 0);
 
     MetricConfiguration() {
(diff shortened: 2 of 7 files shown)


exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricConfiguration.java
```diff
@@ -0,0 +1,98 @@
...
+    /**
+     * Returns a MetricsServiceStub instance used to make RPC calls.
+     *
+     * @return the metrics service stub.
+     */
+    @Nullable
+    public abstract MetricServiceStub getMetricServiceStub();
+
+    public static Builder builder() {
+        return new AutoValue_MetricConfiguration.Builder();
```
zoercai, line 63: AutoValue is still not generating `AutoValue_MetricConfiguration` for me locally 😕

nilebox: I will try to reproduce this on my machine.

exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricConfiguration.java
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    /** Builder for {@link MetricConfiguration}. */
+    @AutoValue.Builder
+    public abstract static class Builder {
+
+        Builder() {}
```
nilebox, line 70: You need to add following setters:
```java
    public abstract static class Builder {

        Builder() {}


        public abstract Builder setCredentials(Credentials credentials);

        public abstract Builder setAddUniqueIdentifier(boolean addUniqueIdentifier);

        public abstract Builder setMetricServiceStub(MetricServiceStub stub);
```
after making this change the compilation worked for me (the EndToEndTest fails, but that's probably unrelated).

nilebox (commented):
See comment about adding missing property setters.

exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricConfiguration.java
nilebox: See the other comment, you are missing some property setters.

exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricConfiguration.java
zoercai: Thank you!

exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricTranslator.java
```diff
@@ -0,0 +1,213 @@
...
+import java.util.Set;
+import java.util.logging.Logger;
+import java.util.stream.Collectors;
+
+public class MetricTranslator {
+
+  private static final Logger logger = Logger.getLogger(MetricTranslator.class.getName());
+
+  static final String DESCRIPTOR_TYPE_URL = "custom.googleapis.com/OpenTelemetry/";
+  static final String UNIQUE_IDENTIFIER_KEY = "opentelemetry_id";
```
nilebox, line 38: Is this key defined somewhere in the spec? I suppose you just ported it from Python?

exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricTranslator.java
```diff
@@ -0,0 +1,213 @@
...
+    if (uniqueIdentifier != null) {
+      metricBuilder.putLabels(UNIQUE_IDENTIFIER_KEY, uniqueIdentifier);
+    }
+    return metricBuilder.build();
+  }
+
+  static MetricDescriptor mapMetricDescriptor(MetricData metric, String uniqueIdentifier) {
+    String instrumentName = metric.getInstrumentationLibraryInfo().getName();
+    MetricDescriptor.Builder builder = MetricDescriptor.newBuilder().setDisplayName(instrumentName)
+        .setType(DESCRIPTOR_TYPE_URL + instrumentName);
```
nilebox, line 78: This adds the `custom.googleapis.com/OpenTelemetry/` prefix to all metric names.
This is a fine default behavior, but I think we should check if the metric is already prefixed.
e.g. if a metric is called `kubernetes.io/container/uptime` (https://cloud.google.com/monitoring/api/metrics_kubernetes), we should leave it as is.

This is how OpenCensus Go exporter is already implemented: https://github.com/census-ecosystem/opencensus-go-exporter-stackdriver/blob/59d068f8d8ff5b653916aa30cdc4e13c7f15d56e/metrics_proto.go#L419
with a hardcode list of known domains: https://github.com/census-ecosystem/opencensus-go-exporter-stackdriver/blob/59d068f8d8ff5b653916aa30cdc4e13c7f15d56e/metrics_proto.go#L48

exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricExporter.java
```diff
@@ -0,0 +1,205 @@
...
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.UUID;
+import java.util.logging.Logger;
+
+public class MetricExporter implements io.opentelemetry.sdk.metrics.export.MetricExporter {
+
+  private static final String PROJECT_NAME_PREFIX = "projects/";
+  private static final long WRITE_INTERVAL_SECOND = 10;
```
nilebox, line 41: In some recent conversations it has been recommended to use 12 seconds as a timeout for Cloud Monitoring. Could you please change to that value?

exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricExporter.java
```diff
@@ -0,0 +1,205 @@
...
+          CreateMetricDescriptorRequest.newBuilder().setName(PROJECT_NAME_PREFIX + projectId)
+              .setMetricDescriptor(descriptor)
+              .build());
+
+      MetricWithLabels updateKey = new MetricWithLabels(descriptor.getType(),
+          metricData.getDescriptor().getConstantLabels());
+
+      // We are expecting one point per MetricData
+      Optional<MetricData.Point> metricPoint = metricData.getPoints().stream()
+          .reduce((first, second) -> second);
```
nilebox, line 129: Can you please explain what you're doing here? It seems that you're just getting the last element from `getPoints`? why not as
```java
metricData.getPoints().get( metricData.getPoints().size() - 1 );
```
?

Also, should we throw an error if size > 1?

Lastly, if there is exactly one data point, then we could just get the first one (and only).
Sorry if I misunderstood this code.

exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricExporter.java
```diff
@@ -0,0 +1,205 @@
...
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.UUID;
+import java.util.logging.Logger;
```
nilebox, line 36: AFAIK standard logger is not configurable, and it's more common to use slf4j with any implementation of log4j, logback etc.
Since this is a library, we want end users to be able to change the format of logs (use JSON, for example).

exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricExporter.java
```diff
@@ -0,0 +1,205 @@
...
+      List<TimeSeries> allTimesSeries) {
+    List<List<TimeSeries>> batches = Lists.partition(allTimesSeries, MAX_BATCH_SIZE);
+    for (List<TimeSeries> timeSeries : batches) {
+      metricServiceClient.createTimeSeries(projectName, new ArrayList<>(timeSeries));
+    }
+  }
+
+  @Override
+  public ResultCode flush() {
+    // TODO (zoe): add support for flush
```
nilebox, line 169: Could you change the TODO to refer to your GitHub handle please?
Also, isn't this a blocker for using the exporter?

exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricTranslator.java
```diff
@@ -0,0 +1,213 @@
...
+  static final Set<Type> GAUGE_TYPES = ImmutableSet.of(NON_MONOTONIC_LONG, NON_MONOTONIC_DOUBLE);
+  static final Set<MetricData.Descriptor.Type> CUMULATIVE_TYPES = ImmutableSet
+      .of(MONOTONIC_LONG, MONOTONIC_DOUBLE);
+  static final Set<MetricData.Descriptor.Type> LONG_TYPES = ImmutableSet.of(NON_MONOTONIC_LONG, MONOTONIC_LONG);
+  static final Set<MetricData.Descriptor.Type> DOUBLE_TYPES = ImmutableSet
+      .of(NON_MONOTONIC_DOUBLE, MONOTONIC_DOUBLE);
+
+  static final Map<String, Map<String, String>> OTEL_TO_GCP_LABELS = ImmutableMap.<String, Map<String, String>>builder()
+      .put("gce_instance", ImmutableMap.<String, String>builder()
+          .put("host.id", "instance_id")
```
nilebox, line 50: nit: Would be nice define keys as constants. Fine to fix later.
Also, ideally they should be defined in the SDK, since they are part of the [Semantic Conventions spec](https://github.com/open-telemetry/opentelemetry-specification/tree/master/specification/resource/semantic_conventions), but that's fine for now.

exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricTranslator.java
```diff
@@ -0,0 +1,213 @@
...
+  }
+
+  static MonitoredResource mapGcpMonitoredResource(Resource resource) {
+    ReadableAttributes attributes = resource.getAttributes();
+    if (attributes.get("cloud.provider") == null ||
+        !attributes.get("cloud.provider").getStringValue().equals("gcp")) {
+      return null;
+    }
+    String resourceType = attributes.get("gcp.resource_type").getStringValue();
+    if (!(resourceType.equalsIgnoreCase("gce_instance") || resourceType.equalsIgnoreCase("gke_container"))) {
```
nilebox, line 113: These should be constants.

exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricTranslator.java
```diff
@@ -0,0 +1,213 @@
...
+    }
+    return builder.build();
+  }
+
+  static String mapAttributeValueToString(AttributeValue value) {
+    switch (value.getType()) {
+      case STRING:
+        return value.getStringValue();
+      case LONG:
+        return value.getLongValue() + "";
```
nilebox, line 133: There should be something like `Long.toString(long)` IIRC which is more accurate and efficient.

exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricTranslator.java
```diff
@@ -0,0 +1,213 @@
...
+      case DOUBLE:
+        return value.getDoubleValue() + "";
+      case BOOLEAN:
+        return value.getBooleanValue() + "";
+      case STRING_ARRAY:
+        return String.join(", ", value.getStringArrayValue());
+      case LONG_ARRAY:
+        return value.getLongArrayValue().stream().map(Object::toString)
```
nilebox, line 141: Should this be `Long::toString` (same below - cleaner to use typed methods)?

exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricTranslator.java
```diff
@@ -0,0 +1,213 @@
...
+    int nanos;
+    if (CUMULATIVE_TYPES.contains(descriptorType)) {
+      if (!lastUpdatedTime.containsKey(updateKey)) {
+        // The aggregation has not reset since the exporter
+        // has started up, so that is the start time
+        seconds = exporterStartTime.getEpochSecond();
+        nanos = exporterStartTime.getNano();
+      } else {
+        // The aggregation reset the last time it was exported
+        // Add 1ms to guarantee there is no overlap from the previous export
```
nilebox, line 197: This seems artificial / misleading, but fine to discuss later if there's a better approach.

nilebox (commented):
Looks good overall, just some clarifying / cleanup comments

exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricTranslator.java
```diff
@@ -0,0 +1,213 @@
...
+  }
+
+  static String mapAttributeValueToString(AttributeValue value) {
+    switch (value.getType()) {
+      case STRING:
+        return value.getStringValue();
+      case LONG:
+        return value.getLongValue() + "";
+      case DOUBLE:
+        return value.getDoubleValue() + "";
```
nilebox, line 135: I think we need to explicitly define the format for converting a double to string, otherwise we may get something like `12e7` (see https://docs.oracle.com/javase/6/docs/api/java/lang/Double.html#toString(double)), i.e. we need something like
```java
String.format("%f", double)
```
or you may look at other suggestions in https://stackoverflow.com/questions/16098046/how-do-i-print-a-double-value-without-scientific-notation-using-java

exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricExporter.java
zoercai: Ah this was from when I didn't know that each metric data would only have one point, and used a stream because `metricData.getPoints()` returns a `Collection`. I've changed it to `metricData.getPoints().iterator().next()`, and added `logger.error()` for `size != 1`

exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricTranslator.java
zoercai: I didn't see anything related in the spec, yep just ported it from Python

build.gradle
```diff
@@ -29,6 +30,7 @@ subprojects {
                 google_cloud_core            : "com.google.cloud:google-cloud-core:${googleCloudVersion}",
                 google_cloud_grpc            : "com.google.api.grpc:grpc-google-cloud-trace-v2:${googleCloudVersion}",
                 google_cloud_trace           : "com.google.cloud:google-cloud-trace:${googleCloudVersion}",
+                google_cloud_monitoring      : "com.google.cloud:google-cloud-monitoring:${cloudMonitoringVersion}",
```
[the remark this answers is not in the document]
zoercai: Yeah I used a separate version for cloud monitoring because the 1.0.2 version didn't have much and didn't work. I've added a separate `cloudTracingVersion` for tracing, kept it to 1.0.2 for this PR though.

exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricExporter.java
zoercai: I've changed this to return `SUCCESS` always based on https://github.com/open-telemetry/opentelemetry-java/blob/master/exporters/otlp/src/main/java/io/opentelemetry/exporters/otlp/OtlpGrpcMetricExporter.java#L136

exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricExporter.java
```diff
@@ -0,0 +1,205 @@
...
+
+  MetricExporter(
+      String projectId,
+      MetricServiceClient client,
+      boolean addUniqueIdentifier
+  ) {
+    this.projectId = projectId;
+    this.metricServiceClient = client;
+    if (addUniqueIdentifier) {
+      this.uniqueIdentifier = UUID.randomUUID().toString().replaceAll("-", "").substring(0, 8);
```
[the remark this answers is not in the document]
james-bebbington: Might be worth checking with @aabmass as I can't remember all the details.

But I recall the main reason for this was to avoid having errors on the Stackdriver side if you send the same metric from two different applications (since Stackdriver doesn't have any equivalent of collection aggregation and gives an error if it receives the same metric timeseries info within 10s). This should not be an issue in most cases as metric data will be sent from different **resources** (this doesn't cover the case of having two processes on the same machine, but for that edge case I think it's reasonable for the user to handle that themselves).

==> I would recommend not implementing this unless we find a good use case for it.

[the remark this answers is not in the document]
nilebox: I agree. @zoercai could you please just remove all usage of "unique identifier" for now?

exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricTranslator.java
```diff
@@ -35,6 +37,8 @@
   private static final Logger logger = LoggerFactory.getLogger(MetricTranslator.class);
 
   static final String DESCRIPTOR_TYPE_URL = "custom.googleapis.com/OpenTelemetry/";
+  static final List<String> KNOWN_DOMAINS = ImmutableList
```
nilebox, line 40: Should this be a `Set<>` for faster lookups?

exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricTranslator.java
```diff
@@ -45,20 +49,28 @@
...
+  static final String GCP_NAMESPACE_KEY = "namespace_id";
+  static final String GCP_POD_KEY = "pod_id";
+  static final String GCP_CONTAINER_KEY = "container_name";
+
   static final Map<String, Map<String, String>> OTEL_TO_GCP_LABELS = ImmutableMap.<String, Map<String, String>>builder()
       .put("gce_instance", ImmutableMap.<String, String>builder()
-          .put("host.id", "instance_id")
-          .put("cloud.account.id", "project_id")
-          .put("cloud.zone", "zone")
+          .put("host.id", GCP_HOST_ID_KEY)
```
nilebox, line 62: Could you also define `OTEL_HOST_ID_KEY` and similar for keys?
They should be defined in the SDK, but for now having a single place with constants would be still useful.

exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricTranslator.java
```diff
@@ -0,0 +1,234 @@
...
+    return DESCRIPTOR_TYPE_URL + instrumentName;
+  }
+
+  static MonitoredResource mapGcpMonitoredResource(Resource resource) {
+    ReadableAttributes attributes = resource.getAttributes();
+    if (attributes.get("cloud.provider") == null ||
+        !attributes.get("cloud.provider").getStringValue().equals("gcp")) {
+      return null;
+    }
+    String resourceType = attributes.get("gcp.resource_type").getStringValue();
```
nilebox, line 133: @james-bebbington @aabmass have we agreed to keep this attribute, or should we switch to [semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/tree/master/specification/resource/semantic_conventions)?

I would suggest to just remove resource GCP mapping for now from the code, and add it later.

nilebox: To clarify, the question was whether we should infer resource type based on other attributes, like we did in https://github.com/open-telemetry/opentelemetry-collector/pull/1462

exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricTranslator.java
nilebox: As mentioned in the other comment, might be better to just remove this mapping code for now, and add it separately later.

exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricTranslator.java
zoercai: Yes! 😁

exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricTranslator.java
```diff
@@ -0,0 +1,143 @@
...
+    long seconds;
+    int nanos;
+    if (CUMULATIVE_TYPES.contains(descriptorType)) {
+      if (!lastUpdatedTime.containsKey(updateKey)) {
+        // The aggregation has not reset since the exporter
+        // has started up, so that is the start time
+        seconds = exporterStartTime.getEpochSecond();
+        nanos = exporterStartTime.getNano();
+      } else {
+        // The aggregation reset the last time it was exported
```
aabmass, line 126: I don't think this comment ported well from Python, since Otel Java doesn't have `stateful`. I'm not too familiar with Otel java but my understanding is that for Counter instruments, `export()` receives a Delta for the interval and so the aggregation is "reset" *every* time (it's different in python because of the `stateful` option we discussed). So for Counters , this branch will actually run every export cycle except for the first one when `lastUpdateTime` is missing the key. I talked to @kjordy and this "fake deltas" approach is fine (so long as GCM doesn't allow DELTAs for custom metrics, which would be better).
```suggestion
        // Send "fake deltas" since GCM doesn't support DELTA for custom metrics
```

However, I would imagine that for `SumObserver`, you would still get `MONOTONIC_*` but then the values that `export()` receives would actually already be cumulative. See [here in the spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/metrics/api.md#sumobserver):
> SumObserver is the asynchronous instrument corresponding to Counter, used to capture a monotonic sum with `Observe(sum)`. "Sum" appears in the name to remind users that it is used to capture sums directly.

For SumObservers then, you would actually want to continue using `exporterStartTime` here, IIUC.

exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricTranslator.java
```diff
@@ -0,0 +1,143 @@
...
+      .of(NON_MONOTONIC_DOUBLE, MONOTONIC_DOUBLE);
+
+  static Metric mapMetric(MetricData metric, String type) {
+    Metric.Builder metricBuilder = Metric.newBuilder().setType(type);
+    metric.getDescriptor().getConstantLabels().forEach(metricBuilder::putLabels);
+    return metricBuilder.build();
+  }
+
+  static MetricDescriptor mapMetricDescriptor(MetricData metric) {
+    String instrumentName = metric.getInstrumentationLibraryInfo().getName();
```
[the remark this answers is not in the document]
nilebox: Agreed, the library info will probably have something like "Spring auto-instrumentation for OpenTelemetry", i.e. nothing to do with the metrics itself

exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricTranslator.java
```diff
@@ -0,0 +1,143 @@
...
+    } else {
+      logger.error("Metric type {} not supported", metricType);
+      return null;
+    }
+    if (LONG_TYPES.contains(metricType)) {
+      builder.setValueType(MetricDescriptor.ValueType.INT64);
+    } else if (DOUBLE_TYPES.contains(metricType)) {
+      builder.setValueType(MetricDescriptor.ValueType.DOUBLE);
+    } else {
+      logger.error("Metric type {} not supported", metricType);
```
nilebox, line 69: You have the same error message here and above, would be nice to make them more specific (make one mention gauge/cumulative and other long/double?)

exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricTranslator.java
zoercai: Thanks for pointing this out! I checked the proto (https://github.com/open-telemetry/opentelemetry-proto/blob/7ae34aaceaeb19ce6e87b2b6ddb542f0104fd72c/opentelemetry/proto/metrics/v1/metrics.proto#L96) and the way https://github.com/open-telemetry/opentelemetry-java/blob/d42d4cd9e2d79c8e5bdc3c43aac5e077f4147794/exporters/otlp/src/main/java/io/opentelemetry/exporters/otlp/MetricAdapter.java#L195 is mapping the interval, and it seems like it's safe to use the start time from the Point, which simplifies things by quite a bit. Please let me know if this doesn't look correct

build.gradle
```diff
@@ -18,23 +18,31 @@ subprojects {
...
+                google_cloud_monitoring_grpc : "com.google.cloud:grpc-google-cloud-monitoring-v3:${cloudMonitoringVersion}",
+                slf4j                        : "org.slf4j:slf4j-api:${slf4jVersion}",
                 opentelemetry_api            : "io.opentelemetry:opentelemetry-api:${openTelemetryVersion}",
                 opentelemetry_sdk            : "io.opentelemetry:opentelemetry-sdk:${openTelemetryVersion}",
                 opentelemetry_operations_java: "com.google.cloud.opentelemetry:exporter-trace:${opentelemetryOperationsVersion}"
         ]
         testLibraries = [
-                junit: "junit:junit:${junitVersion}",
+                junit       : "junit:junit:${junitVersion}",
+                mockito     : "org.mockito:mockito-inline:${mockitoVersion}",
```
nilebox, line 44: I can't find any usage of Mockito in the code? Are you sure you need this?

exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricExporter.java
```diff
@@ -0,0 +1,201 @@
...
+          .setMetricKind(descriptor.getMetricKind())
+          .build());
+    }
+    createTimeSeriesBatch(metricServiceClient, ProjectName.of(projectId), allTimesSeries);
+    return ResultCode.SUCCESS;
+  }
+
+  private static void createTimeSeriesBatch(MetricServiceClient metricServiceClient, ProjectName projectName,
+      List<TimeSeries> allTimesSeries) {
+    List<List<TimeSeries>> batches = Lists.partition(allTimesSeries, MAX_BATCH_SIZE);
```
nilebox, line 153: It's possible that `allTimesSeries.isEmpty()` with the logic above. Will that be handled correctly here?
May just add a check and return.

nilebox (approved):
A few final comments, and it's ready to merge.

build.gradle
zoercai: I meant to add this change in the next test PR but accidentally added it to this one instead, I've removed it here now

exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricTranslator.java
aabmass: I couldn't tell 100% from looking at that code, but that behavior definitely makes sense 👍

exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricExporter.java
```diff
@@ -0,0 +1,204 @@
...
+
+  private static void createTimeSeriesBatch(MetricServiceClient metricServiceClient, ProjectName projectName,
+      List<TimeSeries> allTimesSeries) {
+    if (allTimesSeries.size() == 0) {
+      return;
+    }
+    List<List<TimeSeries>> batches = Lists.partition(allTimesSeries, MAX_BATCH_SIZE);
+    for (List<TimeSeries> timeSeries : batches) {
+      metricServiceClient.createTimeSeries(projectName, new ArrayList<>(timeSeries));
+    }
```
aabmass, line 159: I believe we could do these batch writes in parallel, although we're not doing it in the python exporter either. Maybe this is overkill, since the exporter would be running in its own thread already and with intervals of at least 10 seconds.

@kjordy any thoughts on that?

exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricExporter.java
```diff
@@ -0,0 +1,204 @@
...
+    createTimeSeriesBatch(metricServiceClient, ProjectName.of(projectId), allTimesSeries);
+    return ResultCode.SUCCESS;
+  }
+
+  private static void createTimeSeriesBatch(MetricServiceClient metricServiceClient, ProjectName projectName,
+      List<TimeSeries> allTimesSeries) {
+    if (allTimesSeries.size() == 0) {
+      return;
+    }
```
aabmass, line 155: nit I don't think this special case is necessary

exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricTranslator.java
```diff
@@ -0,0 +1,123 @@
...
+    }
+    pointBuilder.setInterval(mapInterval(point));
+    lastUpdatedTime.put(updateKey, point.getEpochNanos());
+    return pointBuilder.build();
+  }
+
+  static TimeInterval mapInterval(MetricData.Point point) {
+    Timestamp startTime = Timestamp.newBuilder()
+        .setSeconds(point.getStartEpochNanos() / NANO_PER_SECOND)
+        .setNanos((int) (point.getStartEpochNanos() % NANO_PER_SECOND))
```
aabmass, line 115: nit could make a util function for epochNanos -> `Timestamp` for here and below

exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricTranslator.java
```diff
@@ -0,0 +1,123 @@
...
+      logger.error("Type {} not supported", type);
+      return null;
+    }
+    pointBuilder.setInterval(mapInterval(point));
+    lastUpdatedTime.put(updateKey, point.getEpochNanos());
+    return pointBuilder.build();
+  }
+
+  static TimeInterval mapInterval(MetricData.Point point) {
+    Timestamp startTime = Timestamp.newBuilder()
```
aabmass, line 113: I think you may still need to add up to 1ms delta for the "fake deltas"/Counter instruments, to guarantee that the reset intervals are at least 1ms apart. For the SumObserver instruments (real cumulatives), the current code here should be fine since start will just be the same as the previous export pass.

Or maybe a more general approach would be to always subtract 1ms from the end times, which I think would work for both cases? Idk, anyone else have thoughts on this?

aabmass (approved):
LGTM! Just need to work out that 1ms interval spacing

exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricTranslator.java
zoercai: Currently those types (NON_MONOTONIC_LONG, MONOTONIC_DOUBLE, etc) are the only information we get in Java, and it's not possible to tell apart the specific instruments.
I removed the delta part because it seems like https://github.com/open-telemetry/opentelemetry-java/blob/d42d4cd9e2d79c8e5bdc3c43aac5e077f4147794/exporters/otlp/src/main/java/io/opentelemetry/exporters/otlp/MetricAdapter.java#L155 is treating all of them (apart from SUMMARY which I'm ignoring) as cumulative.
WDYT?

exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricExporter.java
zoercai: Ah yep, looks like if `allTimesSeries` is empty the `partition()` method will return an empty list which won't send any request

exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricExporter.java
kjordy: The batches can be done in parallel, but that can be an optimization for the future if needed. It's more important that timeseries are batched to reduce the number of writes which looks good here.

PR #36, opened 2020-09-10
Metrics test

zoercai:
Tests for https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/pull/33
End to end tests here: https://github.com/zoercai/opentelemetry-operations-java/pull/2

### DIFF
--- a/exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricConfiguration.java
+++ b/exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricConfiguration.java
@@ -19,7 +19,7 @@ import javax.annotation.concurrent.Immutable;
 @Immutable
 public abstract class MetricConfiguration {
 
-  private static final String DEFAULT_PROJECT_ID =
+    private static final String DEFAULT_PROJECT_ID =
       Strings.nullToEmpty(ServiceOptions.getDefaultProjectId());
   private static final Duration DEFAULT_DEADLINE = Duration.ofSeconds(10, 0);
 
--- a/exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricConfiguration.java
+++ b/exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricConfiguration.java
@@ -1,24 +1,31 @@
 package com.google.cloud.opentelemetry.metric;
 
+import static java.time.Duration.ZERO;
+
 import com.google.auth.Credentials;
 import com.google.auto.value.AutoValue;
 import com.google.cloud.ServiceOptions;
 import com.google.cloud.monitoring.v3.stub.MetricServiceStub;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
+import java.time.Duration;
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.Immutable;
 
-/** Configurations for {@link MetricExporter}. */
+/**
+ * Configurations for {@link MetricExporter}.
+ */
 @AutoValue
 @Immutable
 public abstract class MetricConfiguration {
 
     private static final String DEFAULT_PROJECT_ID =
-            Strings.nullToEmpty(ServiceOptions.getDefaultProjectId());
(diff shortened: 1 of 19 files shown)


exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricConfiguration.java
```diff
@@ -19,86 +19,88 @@
...
-        }
+    /**
+     * Builds a {@link MetricConfiguration}.
+     *
+     * @return a {@code MetricsConfiguration}.
+     */
+    public MetricConfiguration build() {
+      Preconditions.checkArgument(
+          !Strings.isNullOrEmpty(getProjectId()),
+          "Cannot find a project ID from either configurations or application default.");
```
nilebox, line 100: Shall we return this error message right after the lookup?
Here we should just return a message that project ID is mandatory, I think.

exporters/metrics/src/test/java/com/google/cloud/opentelemetry/metric/FakeData.java
```diff
@@ -0,0 +1,48 @@
...
+      .create("DescriptorName", "Descriptor description", "Unit", Type.MONOTONIC_LONG,
+          someLabels);
+
+  static Attributes someGceAttributes = Attributes.newBuilder()
+      .setAttribute("cloud.account.id", 123)
+      .setAttribute("host.id", "host")
+      .setAttribute("cloud.zone", "US")
+      .setAttribute("cloud.provider", "gcp")
+      .setAttribute("extra_info", "extra")
+      .setAttribute("gcp.resource_type", "gce_instance")
```
nilebox, line 37: We haven't decided if we want to keep this resource attribute (most likely not), could you please remove it?
Fine to keep other GCP attributes, that will be useful once we add support for translating them.

exporters/metrics/src/test/java/com/google/cloud/opentelemetry/metric/MetricExporterTest.java
```diff
@@ -0,0 +1,143 @@
...
+        .create("Descriptor Name", "Descriptor description", "Unit", Type.SUMMARY,
+            someLabels);
+    MetricData metricData = MetricData
+        .create(summaryDescriptor, aGceResource, anInstrumentationLibraryInfo, ImmutableList.of(aLongPoint));
+
+    ResultCode result = exporter.export(ImmutableList.of(metricData));
+    verify(mockClient, times(0)).createMetricDescriptor(any());
+    verify(mockClient, times(0)).createTimeSeries(any(ProjectName.class), any());
+
+    assertEquals(ResultCode.SUCCESS, result);
```
nilebox, line 140: It seems a bit weird to return SUCCESS if all metric data was unsupported and dropped (with a log message)...
I guess it's fine when it was partially successful (only some metrics were dropped).
Could you check if the spec says anything about that?

nilebox (approved):

exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricConfiguration.java
nilebox: As discussed, this is because the "default project ID" is part of the builder: https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/blob/0896e0dce64f836f7587d8d02877dd4e9225fca4/exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricConfiguration.java#L63

So no change is needed, apologies.

exporters/metrics/src/test/java/com/google/cloud/opentelemetry/metric/MetricExporterTest.java
zoercai: Ah that's a really good point. I can't see anything in the spec, the comments for the `ResultCode` say:
```
enum ResultCode {
    /** The export operation finished successfully. */
    SUCCESS,

    /** The export operation finished with an error. */
    FAILURE
}
```
so I think it does make sense to return FAILURE if some metrics were dropped.

PR #37, opened 2020-09-10
Metrics end to end tests

zoercai:


exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricTranslator.java
```diff
@@ -31,6 +31,8 @@
   static final Set<String> KNOWN_DOMAINS = ImmutableSet
       .of("googleapis.com", "kubernetes.io", "istio.io", "knative.dev");
   static final long NANO_PER_SECOND = (long) 1e9;
+  static final String METRIC_DESCRIPTOR_TIME_UNIT = "ns";
+  static final String METRIC_DESCRIPTOR_DESCRIPTION = "OpenTelemetry metric";
```
nilebox, line 35: This seems to say nothing, since we are using OpenTelemetry SDK 🙂 
Why do we need to set it?
If it's required by mock server, we may set it in tests and not in production?

exporters/metrics/src/test/java/com/google/cloud/opentelemetry/metric/EndToEndTest.java
```diff
@@ -0,0 +1,83 @@
...
+public class EndToEndTest {
+
+  private static final String LOCALHOST = "127.0.0.1";
+
+  private Process mockServerProcess;
+  private MetricExporter exporter;
+  private MockCloudMetricClient mockClient;
+
+  @Before
+  public void setup() throws IOException {
```
nilebox, line 38: nit: Assuming that this setup duplicates the one from tracing test, shall we extract it to a base class or some static helper function to be used by both?

exporters/metrics/src/test/java/com/google/cloud/opentelemetry/metric/FakeData.java
```diff
@@ -24,8 +24,11 @@
 
   static Labels someLabels = Labels.newBuilder().setLabel("label1", "value1").setLabel("label2", "False").build();
 
+  // The name does not have to start with "opentelemetry/", it is set this way because of a bug in the mock server,
```
nilebox, line 27: 👍

nilebox (approved):

exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricTranslator.java
zoercai: Yep it's required by the mock server. I've changed this to use `metric.getDescriptor().getDescription()`

exporters/metrics/src/test/java/com/google/cloud/opentelemetry/metric/EndToEndTest.java
zoercai: Since the trace and metric tests are in different modules, I'll leave it for now as we discussed in chat :)

PR #39, opened 2020-09-15
Init: Auto Instrumentation support

AlexanderMann:
# Motivation

Building a jar isn't always an option for third party projects and tools. Leveraging auto-exporting is a really potent tool.

## Suggested Musical Pairing

https://soundcloud.com/thundercat-official/overseas-feat-zack-fox?in=thundercat-official/sets/it-is-what-it-is-23611641

### DIFF
--- a/exporters/auto/README.md
+++ b/exporters/auto/README.md
@@ -1,0 +1,11 @@
+# Auto-Instrumentation for OpenTelemetry in Google Cloud
+
+## Setup
+
+To instrument traces using the `opentelemetry-javaagent`, `opentelemetry-operations-java-auto-<version>.jar` can be used to provide opentelemetry exporters.
+
+```
+java -javaagent:path/to/opentelemetry-javaagent-<version>-all.jar \
+     -Dotel.exporter.jar=path/to/opentelemetry-operations-java-auto-<version>.jar \
+     -jar myapp.jar
+```
--- a/exporters/auto/README.md
+++ b/exporters/auto/README.md
@@ -2,7 +2,7 @@
 
 ## Setup
 
-To instrument traces using the `opentelemetry-javaagent`, `opentelemetry-operations-java-auto-<version>.jar` can be used to provide opentelemetry exporters.
+To instrument metrics and traces using the `opentelemetry-javaagent`, `opentelemetry-operations-java-auto-<version>.jar` can be used to provide opentelemetry exporters.
 
 ```
 java -javaagent:path/to/opentelemetry-javaagent-<version>-all.jar \
--- a/exporters/auto/build.gradle
+++ b/exporters/auto/build.gradle
@@ -1,0 +1,16 @@
+description = 'Auto Exporter for OpenTelemetry'
+
+dependencies {
+    repositories {
+        maven {
+            url "https://dl.bintray.com/open-telemetry/maven"
+        }
+    }
+
+    api(libraries.auto_value_annotations)
+    annotationProcessor(libraries.auto_value)
+    implementation("io.opentelemetry.instrumentation.auto:opentelemetry-javaagent-tooling:0.8.0")
+    implementation(libraries.opentelemetry_sdk)
+    implementation(libraries.opentelemetry_operations_java)
+    implementation(libraries.slf4j)
+}
--- a/exporters/auto/build.gradle
+++ b/exporters/auto/build.gradle
@@ -1,3 +1,7 @@
+plugins {
+    id "com.github.johnrengelman.shadow" version "6.0.0"
+}
+
 description = 'Auto Exporter for OpenTelemetry'
 
 dependencies {
--- a/exporters/auto/build.gradle
+++ b/exporters/auto/build.gradle
@@ -15,6 +15,6 @@ dependencies {
     annotationProcessor(libraries.auto_value)
     implementation("io.opentelemetry.instrumentation.auto:opentelemetry-javaagent-tooling:0.8.0")
     implementation(libraries.opentelemetry_sdk)
-    implementation(libraries.opentelemetry_operations_java)
+    implementation(project(':exporter-trace'))
     implementation(libraries.slf4j)
 }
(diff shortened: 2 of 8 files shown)


AlexanderMann:
Getting traces in our production setup. Pretty happy with how this is working presently. Obviously, there are no tests here, but I'm hoping the maintainers of this repo can offer guidance/pick up some of the work herein.

If you copy the uber jar from CCI you can get this working in your prod env.

exporters/auto/build.gradle
```diff
@@ -0,0 +1,20 @@
...
+dependencies {
+    repositories {
+        maven {
+            url "https://dl.bintray.com/open-telemetry/maven"
+        }
+    }
+
+    api(libraries.auto_value_annotations)
+    annotationProcessor(libraries.auto_value)
+    implementation("io.opentelemetry.instrumentation.auto:opentelemetry-javaagent-tooling:0.8.0")
```
nilebox, line 16: Could you please follow the existing pattern with library names and versions defined in https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/blob/7cac7abd83372548da71a4a6fe463d90b167ffaf/build.gradle#L38-L39?

Most importantly, the `0.8.0` here should be replaced with https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/blob/7cac7abd83372548da71a4a6fe463d90b167ffaf/build.gradle#L24 AFAIU (unless instrumentation repo has its own release cycles - then declare a separate constant).

exporters/auto/README.md
```diff
@@ -0,0 +1,11 @@
+# Auto-Instrumentation for OpenTelemetry in Google Cloud
+
+## Setup
+
+To instrument traces using the `opentelemetry-javaagent`, `opentelemetry-operations-java-auto-<version>.jar` can be used to provide opentelemetry exporters.
+
+```
+java -javaagent:path/to/opentelemetry-javaagent-<version>-all.jar \
+     -Dotel.exporter.jar=path/to/opentelemetry-operations-java-auto-<version>.jar \
```
nilebox, line 9: I'm not familiar with the auto-instrumentation implementation, so I have a question.
This jar currently only implements the trace exporter. If we decided to also implement metrics exporter (is that even supported?), would it be part of the same Gradle project / jar or a separate one?

Having them separately would be more flexible (e.g. a customer may want to only use tracing part of Google Cloud and use a third party vendor for metrics, and vice versa), but requires more boilerplate in the repository.

Since the goal is to provide a simple "auto" mechanism, having one jar for all exporters is probably fine.

nilebox: Update: Yes, metric exporters are supported as well: [MetricExporterFactory](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/master/javaagent-spi/src/main/java/io/opentelemetry/javaagent/spi/exporter/MetricExporterFactory.java)

@AlexanderMann could you implement it as well? Should be very similar to trace exporter.

exporters/auto/src/main/java/com/google/cloud/opentelemetry/auto/GoogleCloudSpanExporterFactory.java
```diff
@@ -0,0 +1,25 @@
...
+import io.opentelemetry.javaagent.tooling.exporter.ExporterConfig;
+import io.opentelemetry.javaagent.tooling.exporter.SpanExporterFactory;
+import io.opentelemetry.sdk.trace.export.SpanExporter;
+import java.io.IOException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class GoogleCloudSpanExporterFactory implements SpanExporterFactory {
+
+    private static final Logger log = LoggerFactory.getLogger(TracerInstaller.class);
```
nilebox, line 14: Why not `GoogleCloudSpanExporterFactory.class`?

exporters/auto/src/main/java/com/google/cloud/opentelemetry/auto/GoogleCloudSpanExporterFactory.java
```diff
@@ -0,0 +1,25 @@
...
+
+    @Override
+    public SpanExporter fromConfig(ExporterConfig _config) {
+        try {
+            return TraceExporter.createWithDefaultConfiguration();
+        } catch (IOException ex) {
+            log.error("Unable to create Google Trace exporter.", ex);
+            return null;
```
nilebox, line 22: There is no check for null in TracerInstaller, and it will throw an NPE in the line 168 AFAIU:
https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/266b8b927d559ac334962756ed8ecfd0effdda5d/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/TracerInstaller.java#L161-L168

Might be better to just wrap exception into a RuntimeException and throw it here?

nilebox (commented):
LGTM overall, see the comments.

exporters/auto/src/main/java/com/google/cloud/opentelemetry/auto/GoogleCloudSpanExporterFactory.java
AlexanderMann: Ah, this is some copy paste from when I implemented this in the OpenTelemetry repo. Will update.

exporters/auto/build.gradle
AlexanderMann: Yup, good call. This was me purely trying to get things working

exporters/auto/README.md
```diff
@@ -0,0 +1,11 @@
+# Auto-Instrumentation for OpenTelemetry in Google Cloud
+
+## Setup
+
+To instrument traces using the `opentelemetry-javaagent`, `opentelemetry-operations-java-auto-<version>.jar` can be used to provide opentelemetry exporters.
```
nilebox, line 5: nit: traces and metrics

nilebox (approved):
LGTM, just one last nit comment.

nilebox (approved):

PR #52, opened 2020-11-09
Start 0.11.0 development cycle

zoercai:


build.gradle
```diff
@@ -5,7 +5,7 @@ subprojects {
     apply plugin: 'signing'
 
     group = "com.google.cloud.opentelemetry"
-    version = "0.10.0-SNAPSHOT" // CURRENT_VERSION
+    version = "0.11.0-SNAPSHOT" // CURRENT_VERSION
```
nilebox, line 8: nit: I think we had a conventional agreement that `-operations-` minor release version would match the OT SDK version, but looks like you've released 0.10.0 which still depends on OT 0.9.1?
https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/blob/17487261f750eb42501d30eec25b6e0c0fe35f71/build.gradle#L26

Not critical until GA since it's mostly a convenience thing.

nilebox (approved):

build.gradle
zoercai: Yeah I realised the mistake right after I released it 😅  I'll make another release under 0.10.1 that depends on 0.10.0 some time

PR #53, opened 2020-11-12
Update to depend on OpenTelemetry 0.10.x

zoercai:


exporters/auto/src/main/java/com/google/cloud/opentelemetry/auto/GoogleCloudMetricExporterFactory.java
```diff
@@ -3,18 +3,25 @@
...
+    } catch (IOException ex) {
+      throw new RuntimeException(ex);
     }
-}
\ No newline at end of file
+  }
+
+  @Override
+  public Set<String> getNames() {
+    return new HashSet<>(Arrays.asList("google_cloud", "google_cloud_metric"));
```
nilebox, line 25: nit: we should probably define these as constants shared between trace and metric exporters
Also, `cloud_monitoring` (and `cloud_trace` respectively) might be more appropriate naming, as these are the names of the products?

nilebox (approved):

PR #55, opened 2020-12-02
Add metric exporter example

zoercai:
This is not working for me on Cloud Monitoring — the metric is not showing up after I run this 😕 

I made https://github.com/zoercai/opentelemetry-operations-java/tree/gce-test to test this on a GCE instance, have also tried running on my Cloudtop instance, as well as locally.


examples/metrics/build.gradle
```diff
@@ -0,0 +1,34 @@
...
+        exclude "META-INF/*.DSA"
+        exclude "META-INF/*.RSA"
+    }
+    with jar
+}
+
+description = 'Examples for Cloud Monitoring Exporter'
+
+dependencies {
+    compile 'io.opentelemetry:opentelemetry-exporters-logging:0.9.1'
```
zoercai, line 29: remove

examples/metrics/src/main/java/com/google/cloud/opentelemetry/example/metrics/MetricsExporterExample.java
```diff
@@ -0,0 +1,67 @@
...
+      OpenTelemetry.getGlobalMeter("instrumentation-library-name", "semver:1.0.0");
+  private static io.opentelemetry.sdk.metrics.export.MetricExporter metricExporter;
+  private static IntervalMetricReader intervalMetricReader;
+
+  public MetricsExporterExample() {
+    try {
+      MetricConfiguration configuration =
+          MetricConfiguration.builder().setDeadline(Duration.ofMillis(30000)).build();
+      metricExporter = MetricExporter.createWithConfiguration(configuration);
+      intervalMetricReader =
```
zoercai, line 27: should this be moved into the metric exporter creation?

nilebox: In the trace example I've made all methods static and extracted into `setupTraceExporter`: https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/blob/master/examples/trace/src/main/java/com/google/cloud/opentelemetry/example/trace/TraceExporterExample.java

I think it would be similar to follow the same structure for metrics?

examples/metrics/src/main/java/com/google/cloud/opentelemetry/example/metrics/MetricsExporterExample.java
```diff
@@ -0,0 +1,67 @@
...
+    } catch (InterruptedException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  public static void main(String[] args) {
+    MetricsExporterExample example = new MetricsExporterExample();
+    example.addCounter();
+    intervalMetricReader.shutdown();
+    metricExporter.flush();
```
nilebox, line 65: Is there an SDK-level shutdown function similar to trace one I used in the example?
https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/blob/7995405f69a9f6d14db6f2e30b9d57653e38a87a/examples/trace/src/main/java/com/google/cloud/opentelemetry/example/trace/TraceExporterExample.java#L74

In that case we wouldn't have to store reference to metricExporter beyong registering it, which would make the code cleaner.

examples/metrics/build.gradle
```diff
@@ -0,0 +1,34 @@
...
+    compile(libraries.opentelemetry_api)
+    compile(libraries.opentelemetry_sdk)
+    compile(libraries.google_cloud_monitoring)
+    compile(libraries.opentelemetry_operations_java_metrics)
```
nilebox, line 33: To follow recent change in trace example, this should be a project dependency.

examples/metrics/src/main/java/com/google/cloud/opentelemetry/example/metrics/MetricsExporterExample.java
zoercai: I don't see anything similar in `OpenTelemetrySdk.getGlobalMeterProvider()`

examples/metrics/src/main/java/com/google/cloud/opentelemetry/example/metrics/MetricsExporterExample.java
```diff
@@ -0,0 +1,71 @@
...
+        METER
+            .longUpDownCounterBuilder("processed_jobs")
+            .setDescription("Processed jobs")
+            .setUnit("1")
+            .build();
+    doWork(counter);
+  }
+
+  private static void doWork(LongUpDownCounter counter) {
+    Random random = new Random();
```
nilebox, line 50: nit: It's better to instantiate random once per program, i.e. make it a `private static final` field.

examples/metrics/src/main/java/com/google/cloud/opentelemetry/example/metrics/MetricsExporterExample.java
```diff
@@ -0,0 +1,71 @@
...
+    try {
+      for (int i = 0; i < 10; i++) {
+        counter.add(random.nextInt());
```
nilebox, line 53: This may lead to the 32-bit integer overflow I think, so it would be safer to use something capped like `random.nextInt(100)`

examples/metrics/src/main/java/com/google/cloud/opentelemetry/example/metrics/MetricsExporterExample.java
```diff
@@ -0,0 +1,74 @@
...
+  private static final Meter METER =
+      OpenTelemetry.getGlobalMeter("instrumentation-library-name", "semver:1.0.0");
+  private static final Random RANDOM = new Random();
+  private static io.opentelemetry.sdk.metrics.export.MetricExporter metricExporter;
+  private static IntervalMetricReader intervalMetricReader;
+
+  private static void setupMetricExporter() {
+    try {
+      MetricConfiguration configuration =
+          MetricConfiguration.builder().setDeadline(Duration.ofMillis(90000)).build();
```
nilebox, line 26: This is a pretty long deadline. Is that because of the issues on your laptop? I'm not sure if we need that long deadline, and example should probably just use `MetricExporter.createWithDefaultConfiguration()`, i.e. the default configuration should be good enough.

examples/metrics/src/main/java/com/google/cloud/opentelemetry/example/metrics/MetricsExporterExample.java
```diff
@@ -0,0 +1,74 @@
...
+              // See https://cloud.google.com/monitoring/quotas#custom_metrics_quotas
+              // Rate at which data can be written to a single time series: one point each 10
+              // seconds
+              .setExportIntervalMillis(20000)
+              .setMetricExporter(metricExporter)
+              .setMetricProducers(
+                  singleton(OpenTelemetrySdk.getGlobalMeterProvider().getMetricProducer()))
+              .build();
+    } catch (IOException e) {
+      e.printStackTrace();
```
nilebox, line 39:
```suggestion
      throw new RuntimeException(e);
```

exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/MetricTranslator.java
```diff
@@ -114,17 +119,41 @@ static Point mapPoint(
...
+          Timestamp.newBuilder()
+              .setSeconds(endTime.getSeconds())
+              .setNanos(endTime.getNanos() + MIN_TIMESTAMP_INTERVAL_NANOS)
+              .build();
+    }
     return TimeInterval.newBuilder().setStartTime(startTime).setEndTime(endTime).build();
   }
 
+  static MonitoredResource mapResource(String projectId) {
+    // This is mapping to the global resource type temporarily:
```
nilebox, line 148: 👍

nilebox (commented):
Just a few final comments

nilebox (approved):

Issue #46, opened 2020-11-03
Unexpected com.google.api.gax.rpc.InvalidArgumentException


christopherL91:
Hi!

I get a strange exception thrown after I've loaded the auto jar, I can see new traces in cloud trace but after a few seconds this pops up:

```
com.google.api.gax.rpc.InvalidArgumentException: io.grpc.StatusRuntimeException: INVALID_ARGUMENT: One or more TimeSeries could not be written: Unrecognized resource type: .: timeSeries[0]
	at com.google.api.gax.rpc.ApiExceptionFactory.createException(ApiExceptionFactory.java:49)
	at com.google.api.gax.grpc.GrpcApiExceptionFactory.create(GrpcApiExceptionFactory.java:72)
	at com.google.api.gax.grpc.GrpcApiExceptionFactory.create(GrpcApiExceptionFactory.java:60)
	at com.google.api.gax.grpc.GrpcExceptionCallable$ExceptionTransformingFuture.onFailure(GrpcExceptionCallable.java:97)
	at com.google.api.core.ApiFutures$1.onFailure(ApiFutures.java:68)
	at com.google.common.util.concurrent.Futures$CallbackListener.run(Futures.java:1039)
	at com.google.common.util.concurrent.DirectExecutor.execute(DirectExecutor.java:30)
	at com.google.common.util.concurrent.AbstractFuture.executeListener(AbstractFuture.java:1165)
	at com.google.common.util.concurrent.AbstractFuture.complete(AbstractFuture.java:958)
	at com.google.common.util.concurrent.AbstractFuture.setException(AbstractFuture.java:749)
	at io.grpc.stub.ClientCalls$GrpcFuture.setException(ClientCalls.java:545)
	at io.grpc.stub.ClientCalls$UnaryStreamToFuture.onClose(ClientCalls.java:515)
	at io.grpc.PartialForwardingClientCallListener.onClose(PartialForwardingClientCallListener.java:39)
	at io.grpc.ForwardingClientCallListener.onClose(ForwardingClientCallListener.java:23)
	at io.grpc.ForwardingClientCallListener$SimpleForwardingClientCallListener.onClose(ForwardingClientCallListener.java:40)
	at io.grpc.internal.CensusStatsModule$StatsClientInterceptor$1$1.onClose(CensusStatsModule.java:700)
	at io.grpc.PartialForwardingClientCallListener.onClose(PartialForwardingClientCallListener.java:39)
[...]

Issue #57, opened 2020-12-03
Question: Warning in the logs on startup of our pods. io.grpc.StatusRuntimeException: INVALID_ARGUMENT: One or more TimeSeries could not be written: Unrecognized resource type: .: timeSeries[0]


dh-ger:
Hi,

We are using the java auto opentelemetry instrumentation along with the cloud trace exporter.
Our dockerfile looks like this:
```
FROM <OUR OPENJDK RUNTIME CONTAINER>

COPY pc-api/build/libs/pc-api-0.0.1-SNAPSHOT.jar /app/pc-api-0.0.1-SNAPSHOT.jar
COPY pc-api/jar/opentelemetry-javaagent-all-v0.10.1.jar /app/opentelemetry-javaagent-all-v0.10.1.jar
COPY pc-api/jar/exporter-auto-0.10.1.jar /app/exporter-auto-0.10.1.jar

ENTRYPOINT ["java","-javaagent:/app/opentelemetry-javaagent-all-v0.11.0.jar",  "-Dotel.exporter.jar=/app/exporter-auto-0.10.1.jar", "-Dotel.propagators=tracecontext,baggage,b3multi,b3", "-jar","/app/pc-api-0.0.1-SNAPSHOT.jar"]

```

We found the following warning in the logs of our pods on startup, and are wondering if it is anything that we should be worried about/if there is any way to get rid of it?

```
[opentelemetry.auto.trace 2020-12-03 15:58:15:087 +0000] [IntervalMetricReader_1] WARN io.opentelemetry.sdk.metrics.export.IntervalMetricReader - Exporter threw an Exception
com.google.api.gax.rpc.InvalidArgumentException: io.grpc.StatusRuntimeException: INVALID_ARGUMENT: One or more TimeSeries could not be written: Unrecognized resource type: .: timeSeries[0]
        at com.google.api.gax.rpc.ApiExceptionFactory.createException(ApiExceptionFactory.java:49)
        at com.google.api.gax.grpc.GrpcApiExceptionFactory.create(GrpcApiExceptionFactory.java:72)
        at com.google.api.gax.grpc.GrpcApiExceptionFactory.create(GrpcApiExceptionFactory.java:60)
        at com.google.api.gax.grpc.GrpcExceptionCallable$ExceptionTransformingFuture.onFailure(GrpcExceptionCallable.java:97)
        at com.google.api.core.ApiFutures$1.onFailure(ApiFutures.java:68)
[...]

Issue #40, opened 2020-09-21
Feature: Fixed Attributes to include common/default env vars provided by GCP and Values set by User


AlexanderMann:
# Motivation

In using auto exporting in production, we're seeing that a bunch of labels Cloud Trace expects to be present are just missing. We would like to be able to see:

- default common env vars populated as their respective labels
- certain env-vars/values passed in turned into labels

## Notes

This is the specific function/abstract method I'm referencing:

https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/blob/7cac7abd83372548da71a4a6fe463d90b167ffaf/exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceConfiguration.java#L59

## Suggested Musical Pairing

https://soundcloud.com/michaelkiwanuka/money

AlexanderMann:
@nilebox I'm hoping you all can provide some insight on _how_ this should be built out or whether I'm just even using this incorrectly 😅. Happy to do the leg work to put this feature in otherwise.

zoercai:
Hi @AlexanderMann! There is currently no mapping of OpenTelemetry attributes to Cloud Trace attributes, which is why you're missing the labels.
The mapping also needs to be added to the metrics exporter for Cloud Monitoring, and since both have the same monitored resource attributes, it would be good to have both trace and metric use a common mapper.
Since there are a lot of attributes (https://cloud.google.com/monitoring/api/resources), I don't think we can add them all at once right now. I can make a skeleton for the common mapper with the basics, and you could add in any additional attributes you need, or I can review your PR if you want to implement the common mapper skeleton as well. Let me know which you'd prefer :)

zoercai:
As for how it should be implemented, the equivalent Go repository has the [metrics exporter mapping implemented](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/master/exporter/metric/metric.go#L317) already, so it would be something similar, with [these changes for the trace exporter](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/84/files)

tomcham:
Hey @zoercai, any updates on having GCP attribute mappings like `opentelemetry-operations-go`? I'm using the metrics exporter and would really appreciate this feature.

dashpole:
[Triage] I think this is still relevant (sorry this has sat for so long).  To confirm, you have resource attributes which you want to attach to all spans?

cc @aabmass 
[...]

PR #58, opened 2020-12-04
Temporarily remove timeseries resource

zoercai:
This will hopefully resolve https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/issues/46 and https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/issues/57

Removing resource from timeseries altogether for now, as it isn't properly mapped. https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/issues/40 will track the resource mapping issue.


exporters/metrics/src/main/java/com.google.cloud.opentelemetry.metric/MetricExporter.java
```diff
@@ -131,18 +129,16 @@ public CompletableResultCode export(Collection<MetricData> metrics) {
...
       Point point = mapPoint(metricData, metricPoint, updateKey, lastUpdatedTime);
       if (point == null) {
         continue;
       }
 
       allTimesSeries.add(
           TimeSeries.newBuilder()
-              .setMetric(metric)
-              .addPoints(point)
-              .setResource(MonitoredResource.newBuilder().build())
```
zoercai, line 144: The lack of the resource here makes the mock server crash, so either the mock server needs to be modified, or the resource mapping needs to be done properly, before the end to end tests can pass.

PR #59, opened 2020-12-04
Fix trace example dependency and fix directory for metric exporter

zoercai:


nilebox (approved):
Thanks!
I will create a follow-up PR improving the trace example code.

PR #60, opened 2020-12-06
Fix and improve the example code

nilebox:
- Use scopes
- Produce child spans
- Invoke `shutdown()` on exit to flush traces


examples/trace/build.gradle
```diff
@@ -29,5 +29,5 @@ dependencies {
     compile(libraries.opentelemetry_api)
     compile(libraries.opentelemetry_sdk)
     compile(libraries.google_cloud_trace)
-    compile(libraries.opentelemetry_operations_java)
```
nilebox, line 32: @zoercai I missed that you already fixed that in https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/pull/59, will rebase if that one gets merged first.

nilebox: Done.

zoercai (approved):

Issue #69, opened 2021-01-22
Better Errors for users when Metrics are submitted in too quick succession or out of order


jsuereth:
Cloud Monitoring can throttle Metric timeseries that are sent in too quick sucession, or out of order.

This client can take two approaches to the problem:

1. Avoid sending the data through detection and issuing a clear error message for users.
2. Handling the failure to write the TimeSeries gracefully and issuing a clear error message for users.

Current approach is to try #1, but we're lacking refinement in the error messages for users to diagnose and fix the issue.

aabmass:
Not prioritizing this right now but would be happy to review a PR based on Josh's proposed solution.

aabmass:
Looking through the backlog, still not a high priority in Java. It would be nice to fix this is in the collector.

Issue #68, opened 2021-01-22
Don't call `createMetricDescriptor` on every `Export` call


jsuereth:
Currently, every export attempts to update the metric descriptor.  This leads to a few problems.

1. It's wasted network bandwidth
2. If we have exports with missing labels we can cause churn in the visible data on google cloud monitoring.

We should adopt a solution simiilar to other SDKs, e.g in JavaScript MetricDescriptors are only registered *once* per Exporter (not export call).  Optionally we could look at inspecting a metric descriptor and doing a good merge.

An additional complication is that we may make assumptions on label-types that are invalid in practice if we don't consider more than one label.  Also, this needs to be done OFF bandwidth on a machine and not take too much time.

Issue #64, opened 2021-01-05
MetricExporter doesn't work with tags


zls88:
I've got some code that can be simplified to something like this:

```
    private final LongCounter callCounter = meter.longCounterBuilder("opentelemetry_responses")
            .setDescription("number of responses returned")
            .setUnit("calls")
            .build();

        @Override
    public void filter(ContainerRequestContext containerRequestContext, ContainerResponseContext containerResponseContext) {
        String statusCode = "" + containerResponseContext.getStatus();
        String methodName = getMethodName(containerRequestContext);

        callCounter.add(1,
                Labels.of(
                        "status_code", statusCode,
                        "method", methodName));
    }
```

I expect it to be able to receive a bunch of combinations of tags and forward them to Stackdriver (presumably with different data). However I actually get an error like this:
```
ERROR [2021-01-05 22:54:01,024] com.google.cloud.opentelemetry.metric.MetricExporter: There should be exactly one point in each metricData, found 0
```
The above was emitted before any data was reported. If I report data, it will emit `There should be exactly one point in each metricData, found <N>` where `<N>` is the number of tag combinations I've reported.

Perhaps we need to iterate over all the points in the loop in `com.google.cloud.opentelemetry.metric.MetricExporter#export` and then perform the rest of the body of that method once per point?

jsuereth:
FYI - Working on a fix for this.

jsuereth:
FYI 0.14.0-RC1 is out (against OTel Java 0.14.1).  @zls88  Would you mind giving this a try and seeing if it fixes things?  I'd like to make sure we have this fixed before the 0.14.0 release if we can.

zls88:
I'll try and test this in the next day or two and get back to you!

jsuereth:
0.14.0-alpha version should fix this issue. Please reopen if you find that isn't the case.

PR #70, opened 2021-01-22
Wip fix for trace 1.0

jsuereth:
- Update readme and build so that failures to run the mock server are super obvious
- Update all APIs to 0.14.1 pending the 1.0 release.
- Overhaul MetricExporter due many changes in the SDK around "type safe point usage" and known issues in the current impl
  - Make room for solving #64 via new "builder" algorithm design.
  - Make room for solving #68 via new design for exporter where metric descriptors are first aggregated before calling create.
  - Remove timestamp throttling pending design on #69


exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/MetricExporter.java
```diff
@@ -96,62 +93,73 @@ private static MetricExporter createWithCredentials(
...
+    // General Algorithm for export:
+    // 1. Iterate over all points in the set of metrics to export
+    // 2. Attempt to register MetricDescriptors if not already registered.
+    // 3. Fire the set of time series off.
+    MetricTimeSeriesBuilder builder =  new AggregateByLabelMetricTimeSeriesBuilder(projectId);
+    for (final MetricData metricData : metrics) {
+      // Extract all the underlying points.
+      switch(metricData.getType()) {
+        case LONG_GAUGE:
+          for(LongPoint point : metricData.getLongGaugeData().getPoints()) {
```
aabmass, line 105: I'm not too familiar with the Java SDK – in practice, is there always just one point?

exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/MetricExporter.java
```diff
@@ -96,62 +93,73 @@ private static MetricExporter createWithCredentials(
...
-
-      // Cloud Monitoring API allows, for any combination of labels and
-      // metric name, one update per WRITE_INTERVAL seconds
-      long pointCollectionTime = metricPoint.getEpochNanos();
-      if (lastUpdatedTime.containsKey(updateKey)
-          && pointCollectionTime
-              <= lastUpdatedTime.get(updateKey) / NANO_PER_SECOND + WRITE_INTERVAL_SECOND) {
-        continue;
+      for (final MetricDescriptor descriptor: builder.getDescriptors()) {
+        // TODO: limit this ONCE per JVM.
```
aabmass, line 132: :+1: 
```suggestion
        // TODO (#68): limit this ONCE per JVM.
```

exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/MetricExporter.java
```diff
@@ -96,62 +93,73 @@ private static MetricExporter createWithCredentials(
...
+      // if (lastUpdatedTime.containsKey(updateKey)
+      //     && pointCollectionTime
+      //         <= lastUpdatedTime.get(updateKey) / NANO_PER_SECOND + WRITE_INTERVAL_SECOND) {
+      //   continue;
+      // }
+    }    
+    List<TimeSeries> series = builder.getTimeSeries();
+    createTimeSeriesBatch(metricServiceClient, ProjectName.of(projectId), series);
+    // TODO: better error reporting.
+    if (series.size() < metrics.size()) {
```
aabmass, line 156: Re above comment about each `MetricData` having a list of points; if that's not the case, this would report failure. Should we just remove the check for now if its not adding anything?

exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/MetricExporter.java
```diff
@@ -96,62 +93,73 @@ private static MetricExporter createWithCredentials(
...
+      // // metric name, one update per WRITE_INTERVAL seconds
+      // long pointCollectionTime = metricPoint.getEpochNanos();
+      // if (lastUpdatedTime.containsKey(updateKey)
+      //     && pointCollectionTime
+      //         <= lastUpdatedTime.get(updateKey) / NANO_PER_SECOND + WRITE_INTERVAL_SECOND) {
+      //   continue;
+      // }
+    }    
+    List<TimeSeries> series = builder.getTimeSeries();
+    createTimeSeriesBatch(metricServiceClient, ProjectName.of(projectId), series);
```
aabmass, line 154: Where do GCM API errors get surfaced? It doesn't look like `createTimeSeriesBatch()` throws. API errors seem like the most reasonable case to `return CompletableResultCode.ofFailure()`

exporters/metrics/src/test/java/com/google/cloud/opentelemetry/metric/EndToEndTest.java
```diff
@@ -26,27 +26,37 @@
...
+      socket.close();
 
-    // Start the mock server. This assumes the binary is present and in $PATH.
-    // Typically, the CI will be the one that curls the binary and adds it to $PATH.
-    String[] cmdArray = new String[] {System.getProperty("mock.server.path"), "-address", address};
-    ProcessBuilder pb = new ProcessBuilder(cmdArray);
-    pb.redirectErrorStream(true);
-    mockServerProcess = pb.start();
+      // Start the mock server. This assumes the binary is present and in $PATH.
+      // Typically, the CI will be the one that curls the binary and adds it to $PATH.
```
aabmass, line 38: This doesn't seem to assume it's in `$PATH`? Is the comment outdated?

exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/EndToEndTest.java
```diff
@@ -35,35 +40,45 @@
...
+  public void setup() throws MockServerStartupException {
+    try {
+      // Find a free port to spin up our server at.
+      ServerSocket socket = new ServerSocket(0);
+      int port = socket.getLocalPort();
+      String address = String.format("%s:%d", LOCALHOST, port);
+      socket.close();
+
+      // Start the mock server. This assumes the binary is present and in $PATH.
+      // Typically, the CI will be the one that curls the binary and adds it to $PATH.
```
aabmass, line 60: Same as above

aabmass (approved):
LGTM just left some nits/questions

exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/MetricExporter.java
```diff
@@ -96,62 +93,73 @@ private static MetricExporter createWithCredentials(
...
+          break;
+        case DOUBLE_SUM:
+          for(DoublePoint point : metricData.getDoubleSumData().getPoints()) {
+            builder.recordPoint(metricData, point);
+          }
+          break;
+        default:
+          logger.error(
+          "Metric type {} not supported. Only gauge and cumulative types are supported.",
+          metricData.getType());
```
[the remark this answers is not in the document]
jsuereth: fixed

exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/MetricExporter.java
jsuereth: I need to refactor all of that code.  I opened up an issue to track it, but yeah... it's not clear what happens right now.

Even more, we may want to use an asynchronous client to avoid blocking, as we return a completable result (i.e an async observable).  We'll fix this up more post-1.0 as it won't break existing usage, just make it better.

exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/MetricExporter.java
jsuereth: I'm trying to not change the behavior right before 1.0.  I agree this needs to get fixed, but I'd rather spend some time on it.

Given metrics is alpha, I think the TODO and a bug to fix is the right way to go for now.

exporters/metrics/src/test/java/com/google/cloud/opentelemetry/metric/EndToEndTest.java
jsuereth: Yeah, I'm going to fix this up in a follow-on.  I want to use test-containers here, maybe we discuss that more offline.

exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/EndToEndTest.java
jsuereth: same comment as above, also I'd like to share this in a test-util library component that's not released.

PR #71, opened 2021-01-25
Improve MetricDescritpor usage in Alpha release of exporter and update documentation.

### LINKED ISSUE #66: tracer README.md still says it isn't on Maven Central
#6 is closed now, but https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/blob/master/exporters/trace/README.md still says that it must be built locally. The text should instead probably be something along the lines of

To add a dependency on GCP trace export using Maven, use the following:
```
<dependency>
    <groupId>com.google.cloud.opentelemetry</groupId>
    <artifactId>exporter-trace</artifactId>
    <version>0.11.0</version>
</dependency>
```

To add the dependency using Gradle:
```
dependencies {
    implementation("com.google.cloud.opentelemetry:exporter-trace:0.11.0")
}
```

### LINKED ISSUE #68: Don't call `createMetricDescriptor` on every `Export` call

jsuereth:
- Add a configurable MetricDescriptor strategy for how/when metric descriptors are sent to GCM
- Update README to point at exporters (auto is pending more testing)
- Update tracing README to denote how to consume maven artifacts.
- Sprinkled a few more TODOs around post 1.0 Tracing tasks.

Fixes #68 
Fixes #66


exporters/metrics/README.md
```diff
@@ -0,0 +1,76 @@
...
+       SdkMeterProvider.builder().buildAndRegisterGlobal()
+     )).build()    
+```
+
+| Configuration | Environment Variable | JVM Property | Description | Default |
+| ------------- | -------------------- | ------------ | ----------- | ------- |
+| projectId     | GOOGLE_CLOUD_PROJECT or GOOGLE_APPLICATION_CREDENTIALS | ??? | The cloud project id.  This is autodiscovered. | The autodiscovered value. |
+| credentials | GOOGLE_APPLICATION_CREDENTIALS | N/A | Credentials to use when talking to Cloud Monitoring API. | App Engine, Cloud Sheel, GCE built-in or provided by `gcloud auth applciation-default login` |
+| deadline      | ??? | ??? | The deadline limit on exprot calls to Cloud Monitoring API | 10 seconds |
+| metricDescriptorStrategy | ??? | ??? | How to adapt OpenTelemetry metric defintiion into google cloud. `ALWAYS_SEND` will try to create metric descriptors on every export.  `SEND_ONCE` will try to create metric descriptors once per Java instance/classloader. `NEVER_SEND` will rely on Cloud Monitoring's auto-generated MetricDescriptors from time series. | `SEND_ONCE` |
```
punya, line 74:
```suggestion
| credentials | GOOGLE_APPLICATION_CREDENTIALS | N/A | Credentials to use when talking to Cloud Monitoring API. | App Engine, Cloud Shell, GCE built-in or provided by `gcloud auth application-default login` |
| deadline      | ??? | ??? | The deadline limit on export calls to Cloud Monitoring API | 10 seconds |
| metricDescriptorStrategy | ??? | ??? | How to adapt OpenTelemetry metric definition into google cloud. `ALWAYS_SEND` will try to create metric descriptors on every export.  `SEND_ONCE` will try to create metric descriptors once per Java instance/classloader. `NEVER_SEND` will rely on Cloud Monitoring's auto-generated MetricDescriptors from time series. | `SEND_ONCE` |
```

exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/MetricConfiguration.java
```diff
@@ -55,10 +55,20 @@
    */
   public abstract Duration getDeadline();
 
+  /**
+   * Returns the strategy for how to send metric descriptors to Cloud Monitoring.
+   * 
+   * <p>The Default is to only send descriptors once per process/classloader.
```
punya, line 61: Do we need to say a bit more about the "per classloader" possibility? E.g., what's the right way to set up the classpath under OSGi or an app server, and what behavior to expect?

(This depends on our audience, which I don't understand very well at the moment.)

exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/MetricDescriptorStrategy.java
```diff
@@ -0,0 +1,61 @@
...
+ */
+public interface MetricDescriptorStrategy {
+    /**
+     * Determines what to do with metrci descriptors.
+     * 
+     * @param batchDescriptors The set of metrics being exported in a batch.
+     * @param client           A consumer that will ensure metric descriptors are
+     *                         registered to cloud monitoring.
+     */
+    void exportDescriptors(Collection<MetricDescriptor> batchDescriptors, Consumer<MetricDescriptor> export);
```
punya, line 21: Could we use `Iterable<MetricDescriptor>` instead of `Collection<...>` here?

exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/MetricDescriptorStrategy.java
```diff
@@ -0,0 +1,61 @@
...
+
+import com.google.api.MetricDescriptor;
+
+import java.util.Collection;
+import java.util.Set;
+import java.util.HashSet;
+import java.util.function.Consumer;
+
+/**
+ * The strategy for how to handle metric descriptors.
```
punya, line 11: Is the intention that users could implement this interface to handle metric descriptors in novel ways (such as filtering based on a prefix)? Or that they should use just the ones we provide? If it's the latter, an `enum` might be more appropriate here.

exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/MetricDescriptorStrategy.java
```diff
@@ -0,0 +1,61 @@
...
+    public static MetricDescriptorStrategy NEVER_SEND = new MetricDescriptorStrategy() {
+        @Override
+        public void exportDescriptors(Collection<MetricDescriptor> batchDescriptors,
+                Consumer<MetricDescriptor> export) {
+        }
+    };
+
+    /** A strategy that sends descriptors once per classloader instance. */
+    public static MetricDescriptorStrategy SEND_ONCE = new MetricDescriptorStrategy() {
+        private Set<String> alreadySent = new HashSet<>();
```
punya, line 47:
```suggestion
        private final Set<String> alreadySent = new HashSet<>();
```

punya (approved):
Overall this looks good barring a few nits.

For the TODO items, is it worth referencing an issue, so that we don't forget them (and can triage them to the right release)?

exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/MetricDescriptorStrategy.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.Collection;
+import java.util.Set;
+import java.util.HashSet;
+import java.util.function.Consumer;
+
+/**
+ * The strategy for how to handle metric descriptors.
+ */
+public interface MetricDescriptorStrategy {
+    /**
```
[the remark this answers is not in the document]
jsuereth: I'm going to defer this, if htat's ok.

(a) I'm using the built-in formatter from VisualStudio for now
(b) Style is NOT enforced on this in anyway
(c) I'm going to add the style-checking plugin with auto-formatter and do a clean sweep in a future CL.  No reason to ever have these comments if we can autoformat (which we can, if we configure it).

exporters/metrics/src/test/java/com/google/cloud/opentelemetry/metric/MetricDescriptorStrategyTest.java
```diff
@@ -0,0 +1,77 @@
...
+
+import com.google.api.MetricDescriptor;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public class MetricDescriptorStrategyTest {
+    @Test
```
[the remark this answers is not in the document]
jsuereth: See other comment.

exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/MetricDescriptorStrategy.java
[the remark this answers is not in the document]
jsuereth: See #73

exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/MetricConfiguration.java
jsuereth: Likely, yes.   It's a really nuanced/hard question that's MOST relevant to auto-instrumentation.  I'm not sure what I'd add here yet, but I'll noodle on it.

exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/MetricDescriptorStrategy.java
jsuereth: Scala habits.... yes we likely can

exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/MetricDescriptorStrategy.java
jsuereth: I'm a fan of giving users flexibility when it's simple to do so.

i.e. this is my "escape hatch" to let folks fix issues in a release by overriding the metric descriptor synthesis/detection we have and do what they need.   I can lock it down to an enum, but we have time to decide on the right thing here, as this is still an "alpha" component.

exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/MetricDescriptorStrategy.java
jsuereth: fixed

exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/MetricDescriptorStrategy.java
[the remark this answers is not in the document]
punya: Agreed, thanks.

exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/MetricDescriptorStrategy.java
punya: I brought up the enum question because the Readme suggests that this will eventually be configurable via environment variable. Is the idea that the env variable will let you pick one of the "existing" implementations, but you need to write custom code if you want to do something unusual? That makes sense.

PR #77, opened 2021-01-26
add header checker lint config

aabmass:
This seems to be redundant with spotlessCheck, but it should work with non-java files as well. I am adding this check to all ops repos.

### DIFF
--- a/.github/header-checker-lint.yml
+++ b/.github/header-checker-lint.yml
@@ -1,0 +1,5 @@
+allowedLicenses:
+  - Apache-2.0
+sourceFileExtensions:
+  - java
+  - sh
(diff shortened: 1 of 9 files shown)


.github/header-checker-lint.yml
```diff
@@ -0,0 +1,5 @@
+allowedLicenses:
+  - Apache-2.0
+sourceFileExtensions:
+  - java
+  - sh
```
aabmass, line 5: @jsuereth do we need copyright headers on `.gradle` and `.sh`? Guidelines just say "all source files", what do you think?

punya: I think it's a good idea to include them.

punya (approved):

jsuereth (approved):

PR #78, opened 2021-01-28
Update to use Nebula-Release plugin

jsuereth:
This should dramatically improve our release instructions (in the near term) to just: 

`./gradlew {snapshot}` for snapshot builds and
`./gradlew -Prelease.veresion={version}` for release

Note: The release plugin will autoamtically create local git tags for you.

Long term, this should also be usable for a completed GHA-based publishing mechanism as used by OTel Java.

TL;DR; on Changes

- Leverage settings.gradle for plugin versioning in one location
- Add Nebula Release plugin to buld
- Move from deprecated `uploadArtifacts` task to `publish`
- Various gradle build jiggery-pokery
- Update all version strings to reflect the 0.14 java usage + prep for a 0.14.0 release of the exporters.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -144,20 +144,6 @@ subprojects {
         archives javadocJar, sourcesJar
     }
 
-    signing {
-        required false
-        // Allow using the GPG agent on linux instead of passwords in a .properties file.
-        if (rootProject.hasProperty('signingUseGpgCmd')) {
-          useGpgCmd()
-        }
-        //sign publishing.publications.mavenJava
-        afterEvaluate {
-            if (publishing.publications.findByName('mavenJava') != null) {
-                sign publishing.publications.mavenJava
-            }
-        }
-    }
-
     javadoc {
         if(JavaVersion.current().isJava9Compatible()) {
             options.addBooleanOption('html5', true)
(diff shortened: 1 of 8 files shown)


build.gradle
```diff
@@ -39,8 +65,10 @@ subprojects {
 
     // Set up java codestyle checking and correction.
     plugins.withId("java") {
-        plugins.apply('checkstyle')
```
punya, line 42: Was this removal intentional?

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceTranslator.java
```diff
@@ -45,8 +45,8 @@
 class TraceTranslator {
```
punya, line 45: Is this Java change (and the other one) supposed to be part of this PR?

build.gradle
jsuereth: yes.  When checkstyle is ACTUALLY run (on release) it fails.   ANy style checks NOT part of spotless should be moved there, we don't need more than one checkstyle (when they do different things)

punya:
I checked out this branch locally and ran `./gradlew snapshot`.

1. The snapshot version was inferred as `0.1.0-SNAPSHOT`. I assume this is because there are no tags yet, just wanted to confirm that it's expected.
2. The Javadoc build failed because of a non-existent `@param` in `MetricDescriptorStrategy`. Unrelated to the core of this change, but perhaps worth fixing in this PR so that the task will succeed.

punya (approved):

PR #81, opened 2021-01-28
Update build + release docs for nebula plugin

jsuereth:
We cut a candidate release with the new process and we have a (verbose) way to cut the first release.

We'll cut down steps in release process as we add more automation.


jsuereth:
> What about the version number in `TraceTranslator.java`?

See #82.  I should have a fix shortly

punya (approved):

PR #85, opened 2021-02-04
Add auto detector and exporter + testing for GCE environment

clamclamyan:
These changes are a java port of the gce detector and exporter from go (https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/detectors/gcp/gce.go)

They should eventually be migrated to the opentelemetry repo (perhaps under https://github.com/open-telemetry/opentelemetry-java/tree/main/sdk-extensions/gcp)


clamclamyan:
@nilebox @jsuereth FYI

detectors/resources/build.gradle
```diff
@@ -0,0 +1,17 @@
...
+    id 'com.github.sherter.google-java-format' version '0.9'
+}
+
+repositories {
+    mavenCentral()
+}
+
+dependencies {
+    testImplementation "com.github.tomakehurst:wiremock-jre8:2.27.2"
+    testCompile group: 'junit', name: 'junit', version: '4.12'
```
nilebox, line 12: Should versions of the dependencies be extracted to the top gradle file (like constants below)?

detectors/resources/src/main/java/com.google.cloud.opentelemetry.detectors/GCEResource.java
```diff
@@ -0,0 +1,63 @@
+package com.google.cloud.opentelemetry.detectors;
```
nilebox, line 1: For some reason, you have a directory named `com.google.cloud.opentelemetry.detectors` containing this file.
Instead, in Java it should be a directory `com` with a subdirectory `google` and so on.
Please fix.

detectors/resources/src/test/java/com/google/cloud/opentelemetry/detectors/GCEResourceTest.java
```diff
@@ -0,0 +1,68 @@
+package com.google.cloud.opentelemetry.detectors;
+
+import static com.github.tomakehurst.wiremock.client.WireMock.*;
+import static org.junit.Assert.*;
```
nilebox, line 4: Wildcard imports have a poor readability (hard to see which classes are imported), could you replace it with individual imports please?
FYI you can disable wildcards imports in IntelliJ: https://www.jetbrains.com/help/idea/creating-and-optimizing-imports.html#disable-wildcard-imports

detectors/resources/src/test/java/com/google/cloud/opentelemetry/detectors/GCEResourceTest.java
```diff
@@ -0,0 +1,68 @@
...
+        get(urlEqualTo(endpointPath))
+            .willReturn(
+                aResponse()
+                    .withStatus(200)
+                    .withHeader("Content-Type", "text/plain")
+                    .withBody(responseBody)));
+  }
+
+  @Test
+  public void testGCEResourceWithEmptyAttributesReturnsEmpty() {
```
nilebox, line 34: As far as I understand, this tests validates behaviour when application is not running on GCP, so the test name should reflect that, e.g. `testGCEResourceNotGCP`.

detectors/resources/src/test/java/com/google/cloud/opentelemetry/detectors/GCEResourceTest.java
```diff
@@ -0,0 +1,68 @@
...
+    GCEResource test = new GCEResource();
+
+    Attributes attr = test.getAttributes();
+
+    assertTrue(attr.isEmpty());
```
nilebox, line 39: Might be useful to leave a comment that it's expected that the official metadata URL is unreachable, and so the result will be empty.

exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/MetricTranslator.java
```diff
@@ -54,6 +59,12 @@
   static final Set<MetricDataType> DOUBLE_TYPES = ImmutableSet.of(DOUBLE_GAUGE, DOUBLE_SUM);
   private static final int MIN_TIMESTAMP_INTERVAL_NANOS = 1000000;
 
+  private static final Map<String, AttributeKey<String>> gceMap =
```
nilebox, line 62: Might be useful to leave a comment with a link to https://cloud.google.com/monitoring/api/resources#tag_gce_instance as the source of truth.

exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/MetricTranslator.java
```diff
@@ -135,10 +146,22 @@ static TimeInterval mapInterval(
     return TimeInterval.newBuilder().setStartTime(startTime).setEndTime(endTime).build();
   }
 
-  static MonitoredResource mapResource(String projectId) {
+  static MonitoredResource mapResource(Attributes attributes, String projectId) {
     // This is mapping to the global resource type temporarily:
     // https://cloud.google.com/monitoring/api/resources#tag_global
```
nilebox, line 151: This comment applies to the original code below, we need to rewrite it, e.g.

```
// First, we try to map to known GCP resources

// GCE: https://cloud.google.com/monitoring/api/resources#tag_gce_instance
...

// If none of the standard resource types apply, use the "global" resource:
// https://cloud.google.com/monitoring/api/resources#tag_global
...
```

exporters/metrics/src/test/java/com/google/cloud/opentelemetry/metric/MetricTranslatorTest.java
```diff
@@ -115,4 +120,38 @@ public void testMapConstantLabelWithLongValueSucceeds() {
...
+            SemanticAttributes.HOST_TYPE, "GCE-instance-type");
+    AttributesBuilder attrBuilder = Attributes.builder();
+    testAttributes.forEach(
+        (key, value) -> {
+          attrBuilder.put(key, value);
+        });
+    Attributes attributes = attrBuilder.build();
+
+    Map<String, String> monitoredResourceMap =
+        MetricTranslator.mapResource(attributes, "GCE_pid").getLabelsMap();
```
nilebox, line 143: You should also add check for `monitoredResource.getType().equals("gce_instance")`

detectors/resources/build.gradle
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+plugins {
+	id 'java'
+	id 'com.github.sherter.google-java-format' version '0.9'
```
punya, line 18: We apply `spotless` globally, so this shouldn't be necessary.

detectors/resources/build.gradle
```diff
@@ -0,0 +1,32 @@
...
+}
+
+repositories {
```
punya, line 21: This should also be handled by the global `build.gradle`.

exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/MetricTranslator.java
```diff
@@ -54,6 +59,13 @@
   static final Set<MetricDataType> DOUBLE_TYPES = ImmutableSet.of(DOUBLE_GAUGE, DOUBLE_SUM);
   private static final int MIN_TIMESTAMP_INTERVAL_NANOS = 1000000;
 
+  // Mapping outlined at https://cloud.google.com/monitoring/api/resources#tag_gce_instance
+  private static final Map<String, AttributeKey<String>> gceMap =
+      Map.of(
```
punya, line 64: This method was added in Java 9, and I believe we're aiming for artifacts to be compatible with Java 8.

(Ideally we should verify this in CI instead of burdening developers with a manual check.)

punya (commented):
I don't have enough context to approve the business logic unfortunately.

detectors/resources/build.gradle
jsuereth: +1

detectors/resources/build.gradle
```diff
@@ -0,0 +1,32 @@
...
+	mavenCentral()
+}
+
+dependencies {
+	compile(libraries.opentelemetry_api)
+	compile(libraries.opentelemetry_sdk)
+	compile(libraries.google_cloud_trace)
```
jsuereth, line 28: I don't think this dependency is used.

detectors/resources/build.gradle
```diff
@@ -0,0 +1,32 @@
...
+	compile(libraries.opentelemetry_semconv)
+	compile(testLibraries.junit)
```
jsuereth, line 30: These (junit + wiremock) should be `testImpelmentation`

detectors/resources/build.gradle
```diff
@@ -0,0 +1,32 @@
...
+	id 'java'
+	id 'com.github.sherter.google-java-format' version '0.9'
+}
+
+repositories {
+	mavenCentral()
+}
+
+dependencies {
+	compile(libraries.opentelemetry_api)
```
jsuereth, line 26: Gradle deprecated "compile" configuration for "implementation".  Please use that.

detectors/resources/src/main/java/com/google/cloud/opentelemetry/detectors/GCPMetadataConfig.java
```diff
@@ -0,0 +1,102 @@
...
+    return getAttribute("instance/name");
+  }
+
+  String getAttribute(String attributeName) {
+    try {
+      URL url = new URL(this.url + attributeName);
+      HttpURLConnection connection = (HttpURLConnection) url.openConnection();
+      connection.setRequestProperty("Metadata-Flavor", "Google");
+      InputStream input = connection.getInputStream();
+      if (connection.getResponseCode() == 200) {
```
jsuereth, line 91: FYI - This doesn't actually mean you got a metadataserver response, some kind of metadata response validation is needed.  E.g. many places DNS squat and hijack non-existent DNS lookups with ad-server homepages.  We have an open bug about this and Verizon FIOS users.

Can you add a TODO to validate that the response actually came from GCP metadata server?

detectors/resources/src/main/java/com/google/cloud/opentelemetry/detectors/GCPMetadataConfig.java
```diff
@@ -0,0 +1,102 @@
...
+import java.net.URL;
+import java.nio.charset.StandardCharsets;
+
+/**
+ * Retrieves Google Cloud project-id and a limited set of instance attributes from Metadata server.
+ *
+ * @see <a href="https://cloud.google.com/compute/docs/storing-retrieving-metadata">
+ *     https://cloud.google.com/compute/docs/storing-retrieving-metadata</a>
+ */
+final class GCPMetadataConfig {
```
jsuereth, line 34: Is this class meant to be called many times in the lifecycle of a Java program?

Right now every call to any method will repeatedly hit the HTTP metadata server endpoint.  It'd be nicer if, instead, you had a method that collected all the components of information, issued ONE failure message if it had issues, and then filled out this MetadataConfig class.


Right now GCPMetadataConfig looks like a static, immutable instance.   Ideally make `GCPMetadataConfig` immutable with one static "fetch" *or* you'd rename all these "getXYZ" to "fetchXYZ" and rename the class to `GCPMetadataService`.

detectors/resources/src/test/java/com/google/cloud/opentelemetry/detectors/GCEResourceTest.java
```diff
@@ -0,0 +1,89 @@
...
+  private final GCEResource testResource = new GCEResource(metadataConfig);
+
+  // Helper method to help stub endpoints
+  private void stubEndpoint(String endpointPath, String responseBody) {
+    stubFor(
+        get(urlEqualTo(endpointPath))
+            .willReturn(
+                aResponse()
+                    .withStatus(200)
+                    .withHeader("Content-Type", "text/plain")
```
jsuereth, line 48: NIt:  AFAIK there's a header we can add here to confirm something came from google cloud metadata.

Can you add: "Metadata-Flavor: Google" here and then require that in your HTTP calling code?

detectors/resources/src/test/java/com/google/cloud/opentelemetry/detectors/GCEResourceTest.java
```diff
@@ -0,0 +1,89 @@
...
+import io.opentelemetry.api.common.Attributes;
+import io.opentelemetry.semconv.trace.attributes.SemanticAttributes;
+import java.util.Map;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public class GCEResourceTest {
```
jsuereth, line 36: Can you add a test where 200 OK is returned from some non GCP dns-hijacked endpoint?

examples/trace/build.gradle
```diff
@@ -45,4 +45,5 @@ dependencies {
 	compile(libraries.opentelemetry_sdk)
 	compile(libraries.google_cloud_trace)
 	compile project(':exporter-trace')
+	compile project(':detector-resources')
```
jsuereth, line 48: Is this a dependency on the exporter or separately consumable thing?

exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/MetricTranslator.java
```diff
@@ -54,6 +59,13 @@
   static final Set<MetricDataType> DOUBLE_TYPES = ImmutableSet.of(DOUBLE_GAUGE, DOUBLE_SUM);
   private static final int MIN_TIMESTAMP_INTERVAL_NANOS = 1000000;
 
+  // Mapping outlined at https://cloud.google.com/monitoring/api/resources#tag_gce_instance
+  private static final Map<String, AttributeKey<String>> gceMap =
```
jsuereth, line 63: This might be more simple (and java-8 friendly) expressed as:
```
private static AttributeKey<String> gceToSemantic(String gce) {
  switch(gce) {
    case "project_id": return SemanticAttributes.CLOUD_ACCOUNT_ID;
    ...
    default: return null;
  }
```
That's byte-code compatible with java 8 (just not source).

exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/AggregateByLabelMetricTimeSeriesBuilder.java
```diff
@@ -86,7 +86,7 @@ public void recordPoint(MetricData metric, DoublePoint point) {
     return TimeSeries.newBuilder()
         .setMetric(mapMetric(labels, descriptor.getType()))
         .setMetricKind(descriptor.getMetricKind())
-        .setResource(mapResource(projectId));
+        .setResource(mapResource(metric.getResource().getAttributes(), projectId));
```
jsuereth, line 89: (MInor) Nit:  Since this is called "mapResource" shouldn't it take the `Resource` and return `MonitoredResource`?

jsuereth (commented):
A couple of early comments.  Meta points:

1. We shoudl figure out where the resource detector belongs.  We've been discussing locally whether this should go upstream.  Ok to put it here for now and we can move it.
2. Don't assume the metadataserver won't be DNS hijacked. I can show you bug reports where it is thanks to local DSL service for folks running their own on-prem exporters.
3. Try ensure failure scenarios happen *early* and *once* and then also log them so users know what we did.  For resource detector i'm not sure what the idea path here is (given the specification), but I do know it's impossible to figure out what went wrong without logs.

detectors/resources/src/main/java/com/google/cloud/opentelemetry/detectors/GCPMetadataConfig.java
```diff
@@ -0,0 +1,102 @@
...
+  String getInstanceName() {
+    return getAttribute("instance/name");
+  }
+
+  String getAttribute(String attributeName) {
+    try {
+      URL url = new URL(this.url + attributeName);
+      HttpURLConnection connection = (HttpURLConnection) url.openConnection();
+      connection.setRequestProperty("Metadata-Flavor", "Google");
+      InputStream input = connection.getInputStream();
```
punya, line 90: Currently, if the response status is not 200, the input stream isn't closed until it gets garbage-collected. Consider moving this statement into the `if` body to fix the leak.

detectors/resources/src/main/java/com/google/cloud/opentelemetry/detectors/GCEResource.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    String zone = metadata.getZone();
+    if (!zone.isEmpty()) {
+      attrBuilders.put(SemanticAttributes.CLOUD_ZONE, zone);
+
+      String[] splitArr = zone.split("-");
+      if (splitArr.length > 2) {
+        attrBuilders.put(
+            SemanticAttributes.CLOUD_REGION, String.join("-", Arrays.copyOfRange(splitArr, 0, 2)));
```
punya, line 57: Consider adding a comment here referencing the documentation for region names, and citing a couple of examples. Otherwise the array splitting and recombining is quite mysterious.
```suggestion
            SemanticAttributes.CLOUD_REGION, splitArr[0] + "-" + splitArr[1]);
```

detectors/resources/src/main/java/com/google/cloud/opentelemetry/detectors/GCPMetadataConfig.java
```diff
@@ -0,0 +1,102 @@
...
+
+  boolean isRunningOnGcp() {
+    return !getProjectId().isEmpty();
+  }
+
+  String getProjectId() {
+    return getAttribute("project/project-id");
+  }
+
+  String getZone() {
```
punya, line 53: For `getZone()` and `getMachineType()`, consider adding an example of the raw value showing why it needs to be massaged.

detectors/resources/src/main/java/com/google/cloud/opentelemetry/detectors/GCPMetadataConfig.java
punya: What's the desired behavior if the metadata endpoint times out? Is there ever a scenario where we want to retry the fetch?

punya: For example, do we want to refresh instance/zone info after live migration? (This should be covered in our ops spec, not necessarily a change to this PR).

nilebox (commented):
Just added a few clarifications to Josh's and Punya's comments.

examples/trace/build.gradle
punya: Even for this PR, it may help to change this to a `runtimeOnly` dependency and add a comment that it's loaded via `ServiceLoader`.

detectors/resources/src/main/java/com/google/cloud/opentelemetry/detectors/GCPMetadataConfig.java
jsuereth: I'll work out the details on the metadata header with the team and describe the use case.  Ideally we can set up an integration test that forces them to exist.  I sent you the internal bug to track details here.

I'm ok with any kind of checking, but the suggestion now is to use the metadata tag.   We'll shore this up, but we'll also need to do a similar fix to other SDKs.    For now, let's use the metadata http header (because it's very easy to check) and offline (and outside the CL) we'll work on other validation mechanisms.

build.gradle
```diff
@@ -134,6 +135,7 @@ subprojects {
 			mockito     : "org.mockito:mockito-inline:${mockitoVersion}",
 			slf4j_simple: "org.slf4j:slf4j-simple:${slf4jVersion}",
 			opentelemetry_sdk_testing: "io.opentelemetry:opentelemetry-sdk-testing:${openTelemetryVersion}",
+			wiremock	: "com.github.tomakehurst:wiremock-jre8:${wiremockVersion}"
```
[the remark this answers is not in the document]
clamclamyan: Might be a GitHub issue. Looks good locally:
Screen Shot 2021-02-05 at 12 22 25 pm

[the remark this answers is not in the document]
nilebox: Probably tabs vs spaces

detectors/resources/src/test/java/com/google/cloud/opentelemetry/detectors/GCEResourceTest.java
jsuereth: It's actually both a request + response header.

(a) We should use it in the request header
(b) For now, I think we use it in the response header and we can open a bug to follow up and make sure this mechanism remains.

Example curl output from a GCE vm:

```
$ curl -v "http://metadata.google.internal/computeMetadata/v1/instance/tags" -H "Metadata-Flavor: Google"
...
< HTTP/1.1 200 OK
< Metadata-Flavor: Google
< Content-Type: application/json
< ETag: {redacted}
< Date: Fri, 05 Feb 2021 14:03:45 GMT
< Server: Metadata Server for VM
< Content-Length: 2
< X-XSS-Protection: 0
< X-Frame-Options: SAMEORIGIN
```

detectors/resources/src/main/java/com/google/cloud/opentelemetry/detectors/GCPMetadataConfig.java
```diff
@@ -0,0 +1,168 @@
...
+    }
+  }
+
+  String getAttribute(String attributeName) {
+    try {
+      URL url = new URL(this.url + attributeName);
+      HttpURLConnection connection = (HttpURLConnection) url.openConnection();
+      connection.setRequestProperty("Metadata-Flavor", "Google");
+      if (connection.getResponseCode() == 200
+          && ("Google").equals(connection.getHeaderField("Metadata-Flavor"))) {
```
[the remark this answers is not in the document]
clamclamyan: Yes (String).equals() is null safe :)

detectors/resources/src/main/java/com/google/cloud/opentelemetry/detectors/GCPMetadataConfig.java
```diff
@@ -0,0 +1,168 @@
...
+  public GCPMetadataConfig(String url) {
+    this.url = url;
+  }
+
+  boolean isRunningOnGcp() {
+    return !getProjectId().isEmpty();
+  }
+
+  String getProjectId() {
+    if (cachedAttributes.containsKey(AttributeEndpoints.projectId.toString())) {
```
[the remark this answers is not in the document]
clamclamyan: Good catch. Changing now

detectors/resources/src/main/java/com/google/cloud/opentelemetry/detectors/GCPMetadataConfig.java
```diff
@@ -0,0 +1,122 @@
...
+    if (val != null) {
+      return val;
+    }
+
+    val = fetchAttribute(attributeName);
+    if (!val.isEmpty()) {
+      cachedAttributes.put(attributeName, val);
+    }
+
+    return val;
```
jsuereth, line 101: Note: You can use Java 8's `computeIfAbsent` here on your cache.

```
return cachedAttributes.computeIfAbsent(attributeName, this::fetchAttribute);
```

Note:  `null` will not be added to the map, so you'd want `fetchAttribute` to return `null` on failures/missing values.

detectors/resources/src/main/java/com/google/cloud/opentelemetry/detectors/GCPMetadataConfig.java
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  String fetchAttribute(String attributeName) {
```
jsuereth, line 104: For methods like this (especially if this is public, which it is) it's a good idea to clearly document failure scenarios (or ask for help if you don't know).

For example, this method swallows all IO exceptions, and returns *empty string* on failure.  It's important to know, as a consumer what "bad" output means here, as an empty string is still a valid string, just not in the context of this method.

My ask:  Please document this method with javadoc.  Something llke:

```
/** Returns the attribute from the MetadataServer or an empty string on any failure. */
```

detectors/resources/build.gradle
```diff
@@ -0,0 +1,24 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+description = 'Google Cloud resource provider for OpenTelemetry'
+
+dependencies {
+	implementation(libraries.opentelemetry_api)
+	implementation(libraries.opentelemetry_sdk)
+	implementation(libraries.google_cloud_trace)
```
jsuereth, line 21: I'm still not convinced that the Resource detector needs a dependency on Google Cloud Trace API.  It especially needs to not be true if we want to contribute this upstream into open telemetry.  Can you comment on why this is needed or what error you see removing it?

jsuereth (commented):
Nice improvements.   A few general java comments.

detectors/resources/build.gradle
clamclamyan: It needs a dependency on Google Guava for com.google.common.base.MoreObjects.firstNonNull. I can instead add this to the root build and pass it down? Would that be a better alternative?

detectors/resources/src/main/java/com/google/cloud/opentelemetry/detectors/GCPMetadataConfig.java
nilebox: I believe both `getAttribute` and `fetchAttribute` methods can be made private.
Adding the comment might still be useful though.

detectors/resources/build.gradle
punya: Could we do the thing that `firstNonNull` does, manually? Code reuse is good, but a low-level library like OT (and its associated ecosystem) is better off avoiding dependencies. For example, we'd want to avoid the scenario where a consumer of OT gets their Guava dependency bumped to a new incompatible version, which breaks their code.

clamclamyan: Yep correct @punya @jsuereth Dependency removed

exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/MetricTranslator.java
```diff
@@ -54,6 +61,17 @@
...
+      Stream.of(
```
nilebox, line 66: I find readability of this code questionable (we could just write a static method which would return a map initialized via multiple `.put(...)` calls instead), but it's fine to improve that in a follow-up PR if needed.

nilebox (approved):

PR #90, opened 2021-02-12
Added auto resource detector and exporter for GKE environment + relevant testing

clamclamyan:


clamclamyan:
FYI @nilebox @jsuereth @punya

detectors/resources/src/main/java/com/google/cloud/opentelemetry/detectors/EnvVarMock.java
```diff
@@ -0,0 +1,31 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.cloud.opentelemetry.detectors;
+
+import java.util.Map;
+
+public class EnvVarMock implements EnvVars {
```
nilebox, line 20: This class can be moved to the `/test/` package I think

detectors/resources/src/test/java/com/google/cloud/opentelemetry/detectors/GKEResourceTest.java
```diff
@@ -0,0 +1,104 @@
...
+            .willReturn(
+                aResponse().withHeader("Metadata-Flavor", "Google").withBody(responseBody)));
+  }
+
+  @Test
+  public void testGKEResourceNotGCP() {
+    GKEResource test = new GKEResource();
+
+    /* The default meta data url is unreachable through testing so getAttributes should not detect a
+    GCP environment, hence returning empty attributes */
```
nilebox, line 56: This is correct for "GKE on GCP" (most common) but there are actually GKE distros to run on other platforms (Anthos), e.g. [GKE on AWS](https://cloud.google.com/anthos/clusters/docs/aws/concepts/architecture) or [GKE on-premise](https://cloud.google.com/anthos/clusters/docs/on-prem/1.3/overview).

Still, this is probably fine to merge as is, and add support for GKE outside GCP later.

examples/metrics/build.gradle
```diff
@@ -15,8 +15,13 @@
  */
 plugins {
 	id 'java'
+	id 'com.google.cloud.tools.jib' version '2.7.1'
 }
 
+// Set jib image path up using Google Container Registry
```
nilebox, line 21:
```suggestion
// Set Docker image path here, e.g. using Google Container Registry or Docker Hub
```

examples/metrics/src/main/java/com/google/cloud/opentelemetry/example/metrics/MetricsExporterExample.java
```diff
@@ -18,17 +18,23 @@
...
 import java.io.IOException;
 import java.util.Random;
+import org.apache.log4j.BasicConfigurator;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class MetricsExporterExample {
+  private static final Logger logger = LoggerFactory.getLogger(MetricsExporterExample.class);
+
+  private static SdkMeterProvider METER_PROVIDER =
```
nilebox, line 34: nit: Ideally these changes should be reviewed and merged separately, as they are not related to adding support for GKE resource.

But given that they are fixing issues with the example it might be okay to merge now and improve further later if needed.

exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/MetricExporter.java
```diff
@@ -167,7 +167,12 @@ public CompletableResultCode export(Collection<MetricData> metrics) {
           continue;
       }
       // Update metric descriptors based on configured strategy.
-      metricDescriptorStrategy.exportDescriptors(builder.getDescriptors(), this::exportDescriptor);
+      try {
```
nilebox, line 170: This change is also not related to adding a GKE resource.
I think there is some issue with trying to create a metric descriptor for a metric which already exists in Cloud Monitoring, but not sure what is the correct way to handle this.

examples/metrics/build.gradle
jsuereth, line 21: You should call out this uses JIB in the README.md, and maybe a link to the gradle plugin docs: https://github.com/GoogleContainerTools/jib/tree/master/jib-gradle-plugin.

Ideally this would have a default setting for folks consuming the example.  Maybe put one in here for now with the doc line you list so folks know to go and change it?

Something generic like:

```
jib.to.image = 'gcr.io/my-gcp-project/my-app'
```

detectors/resources/src/main/resources/META-INF/services/io.opentelemetry.sdk.resources.ResourceProvider
```diff
@@ -1 +1,2 @@
-com.google.cloud.opentelemetry.detectors.GCEResource
\ No newline at end of file
+com.google.cloud.opentelemetry.detectors.GCEResource
+com.google.cloud.opentelemetry.detectors.GKEResource
```
jsuereth, line 2: How do these resource detects work?  Is it "last one wins"? Is it "all of them win"?

TL;DR: how do we ensure GKE beats GCE when it needs to?

jsuereth (changes requested):

detectors/resources/src/main/resources/META-INF/services/io.opentelemetry.sdk.resources.ResourceProvider
nilebox: 1. It's "last write wins" in the current implementation: https://github.com/open-telemetry/opentelemetry-java/blob/bace5dce1f042d76ebbfec3e888f136361d58556/sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java#L182-L183
2. But I'm not sure if the order in the `META-INF` guarantees that classes are loaded in the same order (or if it's deterministic at all)? They are loaded using `ServiceLoader`: https://github.com/open-telemetry/opentelemetry-java/blob/bace5dce1f042d76ebbfec3e888f136361d58556/sdk/common/src/main/java/io/opentelemetry/sdk/resources/Resource.java#L105-L109

For now it shouldn't matter which one wins I think, but if we later add support for `cloud.infrastructure_service` (see https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/cloud.md#cloud), we will need for `gcp_gke` to win over `gcp_gce`.

detectors/resources/src/main/java/com/google/cloud/opentelemetry/detectors/EnvVars.java
```diff
@@ -0,0 +1,29 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.cloud.opentelemetry.detectors;
+
+/** Provides API to fetch environment variables */
+public interface EnvVars {
+  EnvVars DEFAULT_INSTANCE = new EnvVarImpl();
```
punya, line 20: Totally optional, but you can write this as

```java
EnvVars DEFAULT_INSTANCE = System::getenv;
```

and avoid creating a class.

detectors/resources/src/main/java/com/google/cloud/opentelemetry/detectors/GKEResource.java
```diff
@@ -0,0 +1,65 @@
...
+    if (containerName != null && !containerName.isEmpty()) {
+      attrBuilders.put(SemanticAttributes.K8S_CONTAINER_NAME, containerName);
+    }
+
+    String clusterName = metadata.getClusterName();
+    if (clusterName != null && !clusterName.isEmpty()) {
+      attrBuilders.put(SemanticAttributes.K8S_CLUSTER_NAME, clusterName);
+    }
+
+    attrBuilders.putAll(gceAttributes);
```
punya, line 62: Are `gceAttributes` meant to override the four things we pull from env variables above? I.e., would it be just as correct to move this code to line 47?

(You don't have to actually move the code, I'm asking just so we understand and document if there is an expected precedence order.)

detectors/resources/src/main/resources/META-INF/services/io.opentelemetry.sdk.resources.ResourceProvider
punya: Is the SPI going to be frozen after 1.0? I.e., will we be able to add a prioritization mechanism or will we have to couple/merge the GCE and GKE detectors to accomplish this goal?

examples/metrics/src/main/java/com/google/cloud/opentelemetry/example/metrics/MetricsExporterExample.java
```diff
@@ -43,24 +49,24 @@ private static void setupMetricExporter() {
...
       throw new RuntimeException(e);
     }
   }
 
   private static void myUseCase() {
-    LongUpDownCounter counter =
+    LongCounter counter =
         METER
-            .longUpDownCounterBuilder("processed_jobs")
+            .longCounterBuilder("clamclamyan_example_counter")
```
punya, line 62: I would prefer if this username wasn't part of the code.

examples/metrics/src/main/java/com/google/cloud/opentelemetry/example/metrics/MetricsExporterExample.java
```diff
@@ -72,13 +78,30 @@ private static void doWork(LongUpDownCounter counter) {
   }
 
   public static void main(String[] args) throws InterruptedException {
+    System.out.println("Starting the metrics-example application");
+
+    BasicConfigurator.configure();
+    logger.info("Test log message");
```
punya, line 84: Is the logger required? This code only appears to use `System.out.println`.

detectors/resources/src/main/resources/META-INF/services/io.opentelemetry.sdk.resources.ResourceProvider
jsuereth: Not he SPI is not freezing with 1.0 yet, confirmed in the last insturmentation SIG.  I'm going to be making some contributions there, but I think we MAY want to iron out a test to make sure GCE/GKE detectors don't conflict.

@clamclamyan can you just open an Issue for us to figure this out later?  We don't need to hold up the PR on this, I was just curious how you were testing this locally and not running into issues.

nilebox:
The `EndToEndTest` is failing for some reason, which doesn't seem related to changes in this PR:
```
org.testcontainers.containers.ContainerLaunchException: Container startup failed

        Caused by:
        org.rnorth.ducttape.RetryCountExceededException: Retry limit hit with exception
            at org.rnorth.ducttape.unreliables.Unreliables.retryUntilSuccess(Unreliables.java:88)
            at org.testcontainers.containers.GenericContainer.doStart(GenericContainer.java:323)
            ... 45 more

            Caused by:
            org.testcontainers.containers.ContainerLaunchException: Could not create/start container
                at org.testcontainers.containers.GenericContainer.tryStart(GenericContainer.java:497)
                at org.testcontainers.containers.GenericContainer.lambda$doStart$0(GenericContainer.java:325)
                at org.rnorth.ducttape.unreliables.Unreliables.retryUntilSuccess(Unreliables.java:81)
                ... 46 more

                Caused by:
                java.lang.IllegalStateException: Container did not start correctly.
                    at org.testcontainers.containers.GenericContainer.tryStart(GenericContainer.java:435)
                    ... 48 more
```

nilebox (approved):

nilebox:
@jsuereth could you look into the issue with starting a container in EndToEndTest? It blocks this PR from merging.

PR #94, opened 2021-03-04
Bump all version to 1.0/1.0-alpha releases and update APIs/examples f…

jsuereth:
…or latest churn.

- Semantic conventions broken out into `Resource`
- Resource in Java SDK is now funky regarding auto configure vs. default.  Will hold off on any spi.ResourceProvider releases until this is less likely to break
- Metric API rename shuffles
- Trace API rename shuffles
- a couple unstable libraries NOW include -alpha in the version
- Span ids + trace ids now always come from context (see SpandData class).


aabmass (approved):
🚀 LGTM

PR #96, opened 2021-03-05
Fix #88 - Status code mapping and expanded tests.

### LINKED ISSUE #88: OTel Status code UNSET gets mapped to google.rpc.Code CANCELLED
Found status code mapping bugs in our other exporters and I don't think is intentional. We map OTel's `UNSET` status code to google.rpc.Code's [`CANCELLED = 1`](https://github.com/googleapis/googleapis/blob/e28aa8570dc86136fe677202c976f4ee00d77f61/google/rpc/code.proto#L41)

https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/blob/3d26be52ceb199595050f281cefabd3676d1ec27/exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceTranslator.java#L216-L217

This mapping code should probably just use value's from the [proto generated code](http://googleapis.github.io/googleapis/java/all/latest/apidocs/com/google/rpc/Code.html#ABORTED).

jsuereth:
- Update status code mapping to match internal specification for OTLP=>Cloud Trace
- Expand the tests for status code mapping to handle "non ok" statuses.

### DIFF
--- a/exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceTranslator.java
+++ b/exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceTranslator.java
@@ -84,7 +84,10 @@ class TraceTranslator {
             .setTimeEvents(toTimeEventsProto(spanData.getEvents()));
     StatusData status = spanData.getStatus();
     if (status != null) {
-      spanBuilder.setStatus(toStatusProto(status));
+      Status statusProto = toStatusProto(status);
+      if (statusProto != null) {
+        spanBuilder.setStatus(statusProto);
+      }
     }
     long end = spanData.getEndEpochNanos();
     if (end != 0) {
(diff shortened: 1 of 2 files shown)


exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceTranslator.java
```diff
@@ -214,15 +217,19 @@ static Status toStatusProto(StatusData status) {
         statusBuilder.setCode(0);
         break;
       case UNSET:
-        statusBuilder.setCode(1);
-        break;
+        // We do not specify a code in the UNSET case.
+        return null;
       case ERROR:
+        statusBuilder.setCode(2);
```
aabmass, line 223: Can we use the `google.rpc.Code` [generated constant](http://googleapis.github.io/googleapis/java/all/latest/apidocs/com/google/rpc/Code.html#UNKNOWN) here and also on L217 and L231?

exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/TraceTranslatorTest.java
```diff
@@ -205,6 +206,13 @@ public void testToStatusProto() {
 
     // The int representation is 0 for canonical code "OK".
     assertEquals(0, spanStatus.getCode());
+
+    Status failStatus = TraceTranslator.toStatusProto(StatusData.create(StatusCode.ERROR, "FAIL!"));
+    assertEquals(2, failStatus.getCode());
```
aabmass, line 211: Same comment as above about generated constant if possible

aabmass (approved):
Nice!

PR #99, opened 2021-03-19
Update readmes for latest release.

jsuereth:
- Add latest maven release badges.
- Update docs for trace exporter for latest version
- Update RELASING.md for test-container build.

### DIFF
--- a/exporters/trace/README.md
+++ b/exporters/trace/README.md
@@ -1,7 +1,9 @@
 # Cloud Trace Exporter for OpenTelemetry
 
   OpenTelemetry Google Cloud Trace Exporter allows the user to send collected traces to Google Cloud. 
-  
+
+[![Maven Central][maven-image]][maven-url]
+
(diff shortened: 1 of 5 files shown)


exporters/trace/README.md
```diff
@@ -27,7 +29,7 @@ config:
 <dependency>
   <groupId>com.google.cloud.opentelemetry</groupId>
   <artifactId>exporter-trace</artifactId>
-  <version>0.11.0</version>
+  <version>0.14.0</version>
```
aabmass, line 32: Should we put a placeholder so we don't need to bump each time?

aabmass (approved):

exporters/trace/README.md
jsuereth: It's a good suggestion.

TL;DR: not worth the hassle (for users or us).

Longer answer:

- We should give users copy-pastable deps, it's a nice thing.
- I couldn't find any way to auto-update this WITHOUT issuing a commit
- We can probably add this to our automation but it's not the end of the world right now.

PR #109, opened 2021-05-24
Initial implementation of e2e tests

jsuereth:
- Implementation of end-to-end integration testing spec for Trace exporter.
- Cloud Build integration for image + test runs against various environments.
- Scaffold framework for adding new test cases over time, including build + docker image.

### DIFF
--- a/cloudbuild-e2e-gce.yaml
+++ b/cloudbuild-e2e-gce.yaml
@@ -1,0 +1,37 @@
+# Copyright 2021 Google
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+steps:
+  # Wait for the image to exist
+  - name: "docker"
+    id: wait-for-image
+    entrypoint: "sh"
+    timeout: 3m
+    env: ["_TEST_SERVER_IMAGE=${_TEST_SERVER_IMAGE}"]
+    args:
+      - e2e-test-server/wait-for-image.sh
+
+  # Run the test
+  - name: $_TEST_RUNNER_IMAGE
+    id: run-tests-gce
+    dir: /
+    env: ["PROJECT_ID=$PROJECT_ID"]
+    args:
+      - gce
+      - --image=$_TEST_SERVER_IMAGE
+
+logsBucket: gs://opentelemetry-ops-e2e-cloud-build-logs
+substitutions:
+  _TEST_RUNNER_IMAGE: gcr.io/opentelemetry-ops-e2e/opentelemetry-operations-e2e-testing:0.5.23
+  _TEST_SERVER_IMAGE: gcr.io/${PROJECT_ID}/opentelemetry-operations-java-e2e-test-server:${SHORT_SHA}
--- a/e2e-test-server/build.gradle
+++ b/e2e-test-server/build.gradle
@@ -15,8 +15,13 @@
  */
 plugins {
 	id 'java'
+	id 'application'
+	id 'com.github.johnrengelman.shadow'
 }
 
+application {
+	mainClassName = 'com.google.cloud.opentelemetry.endtoend.Server'
+}
 
 description = 'End-To-End integration testing server'
 
@@ -28,3 +33,5 @@ dependencies {
 	compile project(':exporter-trace')
 	runtimeOnly project(':detector-resources')
 }
+
+tasks.build.dependsOn tasks.shadowJar
--- a/e2e-test-server/build.gradle
+++ b/e2e-test-server/build.gradle
@@ -1,0 +1,30 @@
+/*
+ * Copyright 2021 Google
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+plugins {
+	id 'java'
+}
+
+
+description = 'End-To-End integration testing server'
+
+dependencies {
+	compile(libraries.opentelemetry_api)
+	compile(libraries.opentelemetry_sdk)
+	compile(libraries.google_cloud_trace)
+	compile(libraries.google_cloud_pubsub)
+	compile project(':exporter-trace')
+	runtimeOnly project(':detector-resources')
+}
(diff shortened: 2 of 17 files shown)


cloudbuild-e2e-gce.yaml
```diff
@@ -0,0 +1,37 @@
...
+    id: run-tests-gce
+    dir: /
+    env: ["PROJECT_ID=$PROJECT_ID"]
+    args:
+      - gce
+      - --image=$_TEST_SERVER_IMAGE
+
+logsBucket: gs://opentelemetry-ops-e2e-cloud-build-logs
+substitutions:
+  _TEST_RUNNER_IMAGE: gcr.io/opentelemetry-ops-e2e/opentelemetry-operations-e2e-testing:0.5.23
```
punya, line 36: Is it possible to have some automation for updating this version number, which appears in three YAML files?

e2e-test-server/src/main/java/com/google/cloud/opentelemetry/endtoend/ScenarioHandlerManager.java
```diff
@@ -0,0 +1,116 @@
...
+    return withTemporaryTracer(
+        (tracer) -> {
+          Span span =
+              tracer
+                  .spanBuilder("basicTrace")
+                  .setAttribute(Constants.TEST_ID, request.testId())
+                  .startSpan();
+          try {
+            return Response.ok();
+          } finally {
```
punya, line 60: What's the point of this `try...finally`? It seems like there's no way for `Response.ok()` to throw. Is it there to establish a consistent pattern for other tests to follow?

e2e-test-server/src/main/java/com/google/cloud/opentelemetry/endtoend/Server.java
```diff
@@ -0,0 +1,139 @@
...
+import com.google.api.core.ApiFutureCallback;
+import com.google.api.core.ApiFutures;
+import com.google.cloud.pubsub.v1.AckReplyConsumer;
+import com.google.cloud.pubsub.v1.Publisher;
+import com.google.cloud.pubsub.v1.Subscriber;
+import com.google.common.util.concurrent.MoreExecutors;
+import com.google.pubsub.v1.PubsubMessage;
+
+/**
+ * Server implements the "integretation test driver" for this language.
```
punya, line 28: Can we link to the spec for integration test drivers here?

e2e-test-server/src/main/java/com/google/cloud/opentelemetry/endtoend/Server.java
```diff
@@ -0,0 +1,139 @@
...
+
+  /** Starts the subcriber pulling requests. */
+  public void start() {
+    subscriber.startAsync().awaitRunning();
+  }
+
+  /** Closes our subscriptions. */
+  public void close() {
+    if (subscriber != null) {
+      subscriber.stopAsync();
```
punya, line 72: Should this also call `awaitTerminated()`?

punya (approved):

e2e-test-server/src/main/java/com/google/cloud/opentelemetry/endtoend/Server.java
```diff
@@ -0,0 +1,139 @@
...
+
+  /** Execute a scenario based on the incoming message from the test runner. */
+  public void handleMessage(PubsubMessage message, AckReplyConsumer consumer) {
+    if (!message.containsAttributes(Constants.TEST_ID)) {
+      consumer.nack();
+      return;
+    }
+    String testId = message.getAttributesOrDefault(Constants.TEST_ID, "");
+    if (!message.containsAttributes(Constants.SCENARIO)) {
+      respond(testId, Response.invalidArugment("Exepcted attribute \"{SCENARIO}\" is missing"));
```
aabmass, line 109: copypasta artifact from python

```suggestion
      respond(testId, Response.invalidArugment(String.format("Expected attribute \"%s\" is missing", Constants.SCENARIO)));
```

e2e-test-server/src/main/java/com/google/cloud/opentelemetry/endtoend/Constants.java
```diff
@@ -0,0 +1,45 @@
...
+      System.getenv().getOrDefault("REQUEST_SUBSCRIPTION_NAME", "");
+  public static String RESPONSE_TOPIC_NAME =
+      System.getenv().getOrDefault("RESPONSE_TOPIC_NAME", "");
+
+  public static ProjectSubscriptionName getRequestSubscription() {
+    return ProjectSubscriptionName.of(PROJECT_ID, REQUEST_SUBSCRIPTION_NAME);
+  }
+
+  /** @return The pubsub channel we get test requests from. */
+  public static ProjectTopicName getResponseChannel() {
```
aabmass, line 42: tiny nit might be better to call this "topic" as pubsub does

aabmass (approved):
some typos

cloudbuild-e2e-gce.yaml
aabmass: There are a number of options ive thought of
1. I believe Renovate can be configured to do this, but it would need access to our GCR registry i suppose
2. We could use a docker tag and just push new versions of that tag. I like that the explicit version number makes it easy to re-run tests though
3. I can inject the container version in as a substitution from the "Triggers" which are managed by terraform.
4. We can actually put the entire [Build "yaml" into terraform](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/cloudbuild_trigger#build) and remove all of the duplication. My concerns there are 1) if any repo needs extra build steps and 2) its harder to see what's happening with out of source build config

e2e-test-server/build.gradle
```diff
@@ -0,0 +1,37 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+plugins {
+	id 'java'
+	id 'application'
+	id 'com.github.johnrengelman.shadow'
+}
+
+application {
```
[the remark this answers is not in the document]
jsuereth: default is false, but I can add it.

e2e-test-server/src/main/java/com/google/cloud/opentelemetry/endtoend/ScenarioHandlerManager.java
jsuereth: Two (maytbe silly?) reasons:

1. To ensure modifications follow this pattern.
2. The bytecode should instantiate response before ending the span.

cloudbuild-e2e-gce.yaml
jsuereth: Let's open a bug against the integration test work/component to look into providing this.

Simialrly, the "wait for image" script should be shared in some fashion instead of copy-pasta.

e2e-test-server/src/main/java/com/google/cloud/opentelemetry/endtoend/Server.java
jsuereth: When it's public.

e2e-test-server/src/main/java/com/google/cloud/opentelemetry/endtoend/Server.java
```diff
@@ -0,0 +1,143 @@
...
+  public void handleMessage(PubsubMessage message, AckReplyConsumer consumer) {
+    if (!message.containsAttributes(Constants.TEST_ID)) {
+      consumer.nack();
+      return;
+    }
+    String testId = message.getAttributesOrDefault(Constants.TEST_ID, "");
+    if (!message.containsAttributes(Constants.SCENARIO)) {
+      respond(
+          testId,
+          Response.invalidArugment(
```
punya, line 112:
```suggestion
          Response.invalidArgument(
```

PR #122, opened 2021-09-20
Wip bump and exemplars

jsuereth:
- Bump to OTel java 1.6
- Add Distribution/Histogram Support to exporter
- Ensure exemplars (for Distribution) are exported.
- Send non-monotonic sums as Gauge metrics.
- Disable DELTA metrics from being sent to cloud monitoring (before they were sent as cumulatives, which we can reenable once we decide that's how we want all exporters to work).


exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/MetricTranslator.java
```diff
@@ -104,28 +114,69 @@ static MetricDescriptor mapMetricDescriptor(
...
     }
     return builder.build();
   }
 
+  private static void fillHistogramType(
+      DoubleHistogramData histogram, MetricDescriptor.Builder builder) {
+    builder.setValueType(MetricDescriptor.ValueType.DISTRIBUTION);
+    switch (histogram.getAggregationTemporality()) {
+      case DELTA:
+        builder.setMetricKind(MetricDescriptor.MetricKind.DELTA);
```
aabmass, line 151: I don't think delta is supported for custom metrics at all right now (see [here](https://cloud.google.com/monitoring/custom-metrics/creating-metrics#:~:text=metricKind%3A%20The%20default%20metric%20kind%20is%20GAUGE%2C%20but%20if%20you%20specify%20the%20metricKind%20parameter%20of%20the%20TimeSeries%20object%2C%20then%20the%20new%20metric%20has%20that%20kind.%20You%20can%20specify%20only%20GAUGE%20and%20CUMULATIVE%20kinds.), maybe you're aware of this already, and just want the errors to bubble up anyway) 

cc @kjordy

exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/MetricTranslator.java
```diff
@@ -104,28 +114,69 @@ static MetricDescriptor mapMetricDescriptor(
...
+      default:
+        logger.error(
+            "Histogram type {} not supported. Only delta and cumulative types are supported.",
+            histogram);
+        return;
+    }
+  }
+
+  private static void fillSumType(SumData<?> sum, MetricDescriptor.Builder builder) {
+    // TODO: Treat non-monotonic sums as gauges?
```
aabmass, line 165: Hmm looking at examples of up down counter from the spec:

> Example uses for UpDownCounter:
> 
> - the number of active requests
> - the number of items in a queue?

Sound like Gauges to me, but you lose the semantic that these things can be aggregated with Sum() function. How would these work as CUMULATIVE in cloud

exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/MetricTranslator.java
```diff
@@ -104,28 +114,69 @@ static MetricDescriptor mapMetricDescriptor(
...
+    switch (sum.getAggregationTemporality()) {
+      case DELTA:
+        builder.setMetricKind(MetricDescriptor.MetricKind.DELTA);
```
aabmass, line 168: Should these be "pseudo cumulatives" with reset start times for each point? @kjordy

exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/MetricTranslator.java
jsuereth: Ok, these are now dropped.  We can fix this going forward once we finish our metric-export specification for how things map.

exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/MetricTranslator.java
jsuereth: Modified these to be gauges for now.

exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/MetricTranslator.java
jsuereth: If it's alright, let's fix DELTA later, right now OTEL by default isn't making any delta aggregations.

aabmass: I also remember some requirement that the exporter should be able to tell the SDK what temporality it wants–is that still on the table? In that case, we could just ask for cumulatives

exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/MetricTranslator.java
```diff
@@ -104,26 +104,59 @@ static MetricDescriptor mapMetricDescriptor(
...
+  private static MetricDescriptor fillHistogramType(
+      DoubleHistogramData histogram, MetricDescriptor.Builder builder) {
+    builder.setValueType(MetricDescriptor.ValueType.DISTRIBUTION);
+    switch (histogram.getAggregationTemporality()) {
+      case CUMULATIVE:
+        builder.setMetricKind(MetricDescriptor.MetricKind.CUMULATIVE);
+        return builder.build();
+      default:
+        logger.error(
+            "Histogram type {} not supported. Only cumulative types are supported.", histogram);
```
aabmass, line 141: Will this log the DELTA part which is the unsupported bit?

aabmass (approved):

exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/MetricTranslator.java
jsuereth: yes

exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/MetricTranslator.java
jsuereth: i believe that was dropped for this initial version, but let's see how the first-cut language implementations go.

PR #124, opened 2021-10-09
Upgrade to OpenTelemetry 1.7.x

jsuereth:
- IntervalMetricReader is replaced by PeriodicMetricReader
  - No longer need flush/shutdown calls that are specific to IMR.
  - Autoconf module now autoflushes/shutsdown
- Exemplar -> ExemplarData rename
- Prefer using gradle application plugin over custom fatjar (fixes issue where examples started failing to report to google cloud APIs)
- Promote `jib` plugin everywhere for easy distribution to GKE/cloud run.

As of this release
- Exemplars now work in Otel + Cloud Monitoring/Cloud Trace
- Autoconfigure module should be the default for cloud-run/serverless (no explicit flush required)
- DELTA metrics still unsupported
- Still using legacy metric namespaces due to GLOBAL resource default.

### DIFF
--- a/examples/metrics/src/main/java/com/google/cloud/opentelemetry/example/metrics/MetricsExporterExample.java
+++ b/examples/metrics/src/main/java/com/google/cloud/opentelemetry/example/metrics/MetricsExporterExample.java
@@ -20,7 +20,6 @@ import io.opentelemetry.api.metrics.LongCounter;
 import io.opentelemetry.api.metrics.Meter;
 import io.opentelemetry.sdk.metrics.SdkMeterProvider;
 import io.opentelemetry.sdk.metrics.export.PeriodicMetricReader;
-
 import java.io.IOException;
 import java.util.Random;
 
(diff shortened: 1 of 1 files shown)


examples/metrics/src/main/java/com/google/cloud/opentelemetry/example/metrics/MetricsExporterExample.java
```diff
@@ -15,39 +15,34 @@
...
-      metricExporter = MetricExporter.createWithDefaultConfiguration();
-      intervalMetricReader =
-          IntervalMetricReader.builder()
-              // See https://cloud.google.com/monitoring/quotas#custom_metrics_quotas
-              // Rate at which data can be written to a single time series: one point each 10
-              // seconds.
-              .setExportIntervalMillis(20000)
-              .setMetricExporter(metricExporter)
-              .setMetricProducers(singleton(METER_PROVIDER))
+      io.opentelemetry.sdk.metrics.export.MetricExporter metricExporter =
```
punya, line 33: Why do we use fully qualified names here instead of importing? It makes the code a bit harder to read (and made me wonder if there were name collisions I had to worry about).

jsuereth: probably just a VisualStudio thing.  Will fix.

PR #132, opened 2022-01-13
Bump otel version and clean up some resource detection

jsuereth:
- Bump to stable metrics API and alter examples/code for it.
- Update resource detection for more semantic conventions
- Add example that just dumps resource detection output in various environments.


dashpole (approved):

detectors/resources/src/main/java/com/google/cloud/opentelemetry/detectors/GKEResource.java
```diff
@@ -49,9 +49,16 @@ public Attributes getAttributes() {
     AttributesBuilder attrBuilders = Attributes.builder();
     attrBuilders.putAll(gceAttributes);
 
-    attrBuilders.put(ResourceAttributes.K8S_NAMESPACE_NAME, envVars.get("NAMESPACE"));
+    attrBuilders.put(
+        ResourceAttributes.CLOUD_PLATFORM,
+        ResourceAttributes.CloudPlatformValues.GCP_KUBERNETES_ENGINE);
     attrBuilders.put(ResourceAttributes.K8S_POD_NAME, envVars.get("HOSTNAME"));
 
+    String namespace = envVars.get("NAMESPACE");
```
aabmass, line 57: In the Python resource detector, we have a fallback which reads `/var/run/secrets/kubernetes.io/serviceaccount/namespace`. I don't remember the history of that, maybe @dashpole knows if this is a good idea or we should rely on the envvar alone.

https://github.com/GoogleCloudPlatform/opentelemetry-operations-python/blob/93cac30c8cf71777a26b1eb4203cd7b1b1472941/opentelemetry-resourcedetector-gcp/src/opentelemetry/resourcedetector/gcp_resource_detector/__init__.py#L75-L82

detectors/resources/src/main/java/com/google/cloud/opentelemetry/detectors/GKEResource.java
```diff
@@ -49,9 +49,16 @@ public Attributes getAttributes() {
     AttributesBuilder attrBuilders = Attributes.builder();
     attrBuilders.putAll(gceAttributes);
 
-    attrBuilders.put(ResourceAttributes.K8S_NAMESPACE_NAME, envVars.get("NAMESPACE"));
+    attrBuilders.put(
+        ResourceAttributes.CLOUD_PLATFORM,
+        ResourceAttributes.CloudPlatformValues.GCP_KUBERNETES_ENGINE);
     attrBuilders.put(ResourceAttributes.K8S_POD_NAME, envVars.get("HOSTNAME"));
```
aabmass, line 55: IIRC, this will be truncated if the pod name is too long. The python detector recommends setting this also via downward API. Wdyt?

aabmass (approved):

detectors/resources/src/main/java/com/google/cloud/opentelemetry/detectors/GKEResource.java
jsuereth: Added this as an optional thing, updated tests.

detectors/resources/src/main/java/com/google/cloud/opentelemetry/detectors/GKEResource.java
jsuereth: Interesting.  I'd like to see what OTEL is doing for general k8s detection outside of our detectors.    I updated the example to include ALL detectors from Java's SDK-extension *plus* ours.

here's the result:

```
{
  cloud.account.id="test-java-gke2", 
  cloud.availability_zone="us-central1-c", 
  cloud.platform="gcp_kubernetes_engine", 
  cloud.provider="gcp", 
  cloud.region="us-central1", 
  container.id="c9a79cde0bf08ad7b7f527fa116b7a35e328242982f23d7b7d2e25fb0b3aab62", 
  host.arch="amd64", 
  host.id="7692099607857750017", 
  host.name="hello-resource-java-8h2cq", 
  k8s.cluster.name="autopilot-cluster-1", 
  k8s.container.name="hello-resource-java",
  k8s.namespace.name="default", 
  k8s.pod.name="hello-resource-java-8h2cq", 
  os.description="Linux 5.4.144+", 
  os.type="linux", 
  process.command_line="/usr/lib/jvm/java-11-openjdk-amd64:bin:java", 
  process.executable.path="/usr/lib/jvm/java-11-openjdk-amd64:bin:java", 
  process.pid=1, 
  process.runtime.description="Debian OpenJDK 64-Bit Server VM 11.0.13+8-post-Debian-1deb10u1", 
  process.runtime.name="OpenJDK Runtime Environment", 
  process.runtime.version="11.0.13+8-post-Debian-1deb10u1", 
  service.name="hello-resource-java", 
  telemetry.sdk.language="java", 
  telemetry.sdk.name="opentelemetry", 
  telemetry.sdk.version="1.10.0"
}
```

jsuereth:
Merging as-is and we can do further resource-detection cleanup rounds later.

PR #134, opened 2022-01-18
Update resource mapping

jsuereth:
This is a breaking change (and will need to be documented/major version bump), but we already have a major version bump coming for OTEL 1.16 change.

- Update resource mapping to be in-line w/ newly written collector resource mapping
- adds expanded tests for `aws_ec2_instance` and `generic_task`
- Never uses `global` resource
- Requires `cloud.platform` semconv attribute to use anything but generic_task.

### DIFF
--- a/exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/AggregateByLabelMetricTimeSeriesBuilder.java
+++ b/exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/AggregateByLabelMetricTimeSeriesBuilder.java
@@ -19,7 +19,7 @@ import static com.google.cloud.opentelemetry.metric.MetricTranslator.mapDistribu
 import static com.google.cloud.opentelemetry.metric.MetricTranslator.mapInterval;
 import static com.google.cloud.opentelemetry.metric.MetricTranslator.mapMetric;
 import static com.google.cloud.opentelemetry.metric.MetricTranslator.mapMetricDescriptor;
-import static com.google.cloud.opentelemetry.metric.MetricTranslator.mapResource;
+import static com.google.cloud.opentelemetry.metric.ResourceTranslator.mapResource;
 
 import com.google.api.MetricDescriptor;
 import com.google.cloud.opentelemetry.metric.MetricExporter.MetricWithLabels;
@@ -106,7 +106,7 @@ public class AggregateByLabelMetricTimeSeriesBuilder implements MetricTimeSeries
     return TimeSeries.newBuilder()
         .setMetric(mapMetric(attributes, descriptor.getType()))
         .setMetricKind(descriptor.getMetricKind())
-        .setResource(mapResource(metric.getResource(), projectId));
+        .setResource(mapResource(metric.getResource()));
   }
 
   @Override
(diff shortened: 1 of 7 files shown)


exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/AggregateByLabelMetricTimeSeriesBuilder.java
```diff
@@ -106,7 +106,7 @@ public void recordPoint(MetricData metric, DoubleHistogramPointData point) {
     return TimeSeries.newBuilder()
         .setMetric(mapMetric(attributes, descriptor.getType()))
         .setMetricKind(descriptor.getMetricKind())
-        .setResource(mapResource(metric.getResource(), projectId));
```
dashpole, line 109: is project Id no longer required?

dashpole (approved):
lgtm overall

exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/AggregateByLabelMetricTimeSeriesBuilder.java
jsuereth: It's a good question.  I did not see us using it in the exporter resources, so I assumed this was fine to do.  Maybe @aabmass knows more

aabmass: It is implicitly the same as the `name` field in the request if omitted. If you do include it, it must match name anyway. So the decision was to just omit it

exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/ResourceTranslator.java
```diff
@@ -0,0 +1,122 @@
...
+          AttributeMapping.create("namespace", ResourceAttributes.SERVICE_NAMESPACE),
+          AttributeMapping.create("job", ResourceAttributes.SERVICE_NAME),
+          AttributeMapping.create("task_id", ResourceAttributes.SERVICE_INSTANCE_ID));
+
+  /** Converts a Java OpenTelemetyr SDK resoruce into a MonitoredResource from GCP. */
+  public static MonitoredResource mapResource(Resource resource) {
+    String platform = resource.getAttribute(ResourceAttributes.CLOUD_PLATFORM);
+    if (platform == null) {
+      return mapBase(resource, "generic_task", GENERIC_TASK_LABELS);
+    }
```
aabmass, line 99: Do we need both this and the `default` case on 108?

aabmass (approved):

dashpole (approved):

exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/ResourceTranslator.java
jsuereth: Yeah. https://docs.oracle.com/javase/specs/jls/se8/html/jls-14.html#jls-14.11

You can't switch on `null` in Java, so you have to filter it out ahead of time or get NPE

PR #143, opened 2022-03-09
Upgrade to OTel Java 1.11.0 and instrumentation 1.11.1

aabmass:
These are the most recent compatible versions (OTel SDK 1.12.0 is out but not supported by the agent yet).


punya (approved):
I think this is okay. The purist position would be to create all-new implementations of those interface in our tests so that we can avoid a dependency on the `internal` packages. I think that's a poor use of our time; we should defer creating such tests until the internal ones break or disappear.

build.gradle
```diff
@@ -109,8 +109,8 @@ subprojects {
 		googleCloudVersion = '2.0.5'
 		googleTraceVersion = '2.0.0'
 		cloudMonitoringVersion = '3.0.0'
-		openTelemetryVersion = '1.10.0'
-		openTelemetryInstrumentationVersion = '1.10.0'
+		openTelemetryVersion = '1.12.0'
```
aabmass, line 112: Ah, checking the tagged released of the auto jar, I think it was released for `1.11.0` so will need to chagne this before merge

punya (approved):
Nice catch about the agent.

Issue #150, opened 2022-05-02
Issue with resolving dependencies


zAPFy:
```
<groupId>com.google.cloud.opentelemetry</groupId>
<artifactId>exporter-trace</artifactId>
<version>0.21.0</version>
```

has a dependecie on 

```
<groupId>com.google.cloud.opentelemetry</groupId>
<artifactId>shared-resourcemapping</artifactId>
<version>0.21.0</version>
<scope>runtime</scope>
```

but no such version exists (there is a alpha version)

PS. not sure why alpha code is included in production grade package but that is a different topic

jsuereth:
Thanks for the report, will get this shored up ASAP.

A note on why RESOURCEMAPPING is alpha (and OTEL versioning)

- Our Exporters rely on "Semantic Conventions" of OTEL.   We do some automatic mapping of OTEL attributes -> GCP labels.   Previously, this may have been done with raw srings, we now rely on an the `-alpha` component.
- The `-alpha` is because the component relies on Java's Semantic Conventions, which are still marked unstable.
- While we're not happy with having this unstable requirement, the BENEFIT is that you, as a user, will (hopefully) get some kind of dependency error if you attempt to use instrumentation that does not use the *SAME* semantic conventions.   Until either stability *or* evolution APIs are created, this is a bit of an issue.

Sorry for the breakages.  In the meantime, you can workaround this by explicitly excluding that dependency and including the proper one:

```xml
<dependency>
  <groupId>com.google.cloud.opentelemetry</groupId>
  <artifactId>exporter-trace</artifactId>
  <version>0.21.0</version>
  <exclusions>
    <exclusion>
      <groupId>com.google.cloud.opentelemetry</groupId>
      <artifactId>exporter-trace</artifactId>
    </exclusion>
  </exclusions>
</dependency>
<dependency>
  <groupId>com.google.cloud.opentelemetry</groupId>
  <artifactId>shared-resourcemapping</artifactId>
  <version>0.21.0-alpha</version>
</dependency>
```

PR #151, opened 2022-05-02
Shade all unstable dependencies used in ResourceMapping.

jsuereth:
- Towards a fix for #150
- We can push a non-alpha (stable) version of this library.


aabmass (approved):
LGTM

Issue #174, opened 2022-11-08
Combine all the GCP Resource Detectors into a single resource detector


psx95:
Currently we have different classes for each GCP compute platform - we need to combine it all into a single class - `GCPResource`.

This simplifies the sdk configuration for the end user where they do not have to explicitly list each of the platform.

PR #176, opened 2022-11-15
Pr/enable gae resource detection

psx95:
### Description

This PR makes the following changes - 
 - Adds resource detection capabilities for GAE (Google App Engine) environments. 
 - Adds steps for cloudbuild in YAML to support automated test runs (_In order to enable the trigger properly, a version update to the e2e-testing image will be required_).

### Testing 
 - Built a docker image for the instrumented test server manually provided it to the e2e test runner, testing it specifically with the GAE arguments. 
      - All tests passed. 

### Notes on GitHub checks - 
 - ~~`ops-java-e2e-gae` will pass once [this PR](https://github.com/GoogleCloudPlatform/opentelemetry-operations-e2e-testing/pull/17) is merged and the new version of e2e-test-runner is used in `cloudbuild-e2e-gae.yaml`.~~  PR is merged.
 -  ~~All checks except `ops-java-e2e-gae` should pass.~~ All checks should pass now.

### DIFF
--- a/detectors/resources/src/main/java/com/google/cloud/opentelemetry/detectors/GAEResource.java
+++ b/detectors/resources/src/main/java/com/google/cloud/opentelemetry/detectors/GAEResource.java
@@ -1,0 +1,77 @@
+/*
+ * Copyright 2022 Google
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.cloud.opentelemetry.detectors;
+
+import io.opentelemetry.api.common.Attributes;
+import io.opentelemetry.api.common.AttributesBuilder;
+import io.opentelemetry.sdk.autoconfigure.spi.ConfigProperties;
+import io.opentelemetry.sdk.autoconfigure.spi.ResourceProvider;
+import io.opentelemetry.sdk.resources.Resource;
+import io.opentelemetry.semconv.resource.attributes.ResourceAttributes;
+
+public class GAEResource implements ResourceProvider {
+  private final GCPMetadataConfig metadata;
+  private final EnvVars envVars;
+
+  public GAEResource() {
+    this.metadata = GCPMetadataConfig.DEFAULT_INSTANCE;
+    this.envVars = EnvVars.DEFAULT_INSTANCE;
+  }
+
+  // For testing only
+  GAEResource(GCPMetadataConfig metadata, EnvVars envVars) {
+    this.metadata = metadata;
+    this.envVars = envVars;
+  }
+
+  public Attributes getAttributes() {
+    if (!metadata.isRunningOnGcp()) {
+      return Attributes.empty();
+    }
+
+    AttributesBuilder attrBuilders = Attributes.builder();
+    attrBuilders.put(ResourceAttributes.CLOUD_PROVIDER, ResourceAttributes.CloudProviderValues.GCP);
+
+    if (envVars.get("GAE_SERVICE") != null) {
+      // add the resource attributes for App Engine
+      attrBuilders.put(
+          ResourceAttributes.CLOUD_PLATFORM, ResourceAttributes.CloudPlatformValues.GCP_APP_ENGINE);
+
+      String appModuleName = envVars.get("GAE_SERVICE");
+      if (appModuleName != null) {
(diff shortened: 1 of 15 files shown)


detectors/resources/src/main/java/com/google/cloud/opentelemetry/detectors/GAEResource.java
```diff
@@ -0,0 +1,77 @@
...
+      if (appVersionId != null) {
+        attrBuilders.put(ResourceAttributes.FAAS_VERSION, appVersionId);
+      }
+
+      String appInstanceId = envVars.get("GAE_INSTANCE");
+      if (appInstanceId != null) {
+        attrBuilders.put(ResourceAttributes.FAAS_ID, appInstanceId);
+      }
+
+      AttributesExtractorUtil.addCloudRegionFromMetadata(attrBuilders, metadata);
```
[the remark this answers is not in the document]
psx95: I see this https://cloud.google.com/appengine/docs/legacy/standard/java/accessing-instance-metadata - which is metadata access on standard GAE, but I am unable to find the similar documentation for flex environment, from where did you get this information ?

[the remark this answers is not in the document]
dashpole: Ran into this when implementing the go detector: https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/408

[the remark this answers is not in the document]
psx95: I see, this is a good catch. 
From your comment above, I think the behavior is reverse - on flex, `instance/zone` works and you have to parse out the region from there. However, from a short experiment I ran on GAE standard, this approach will fail there - 
`instance/zone` returns values like `us-16` from which region cannot be parsed.

I will update this PR to handle the standard case as well (`instance/region` works there), although it will have no affect on our tests since the instrumented test runner runs in a flex container.

detectors/resources/src/main/java/com/google/cloud/opentelemetry/detectors/GAEResource.java
```diff
@@ -0,0 +1,92 @@
...
+    this.envVars = envVars;
+  }
+
+  public Attributes getAttributes() {
+    if (!metadata.isRunningOnGcp()) {
+      return Attributes.empty();
+    }
+
+    AttributesBuilder attrBuilders = Attributes.builder();
+    attrBuilders.put(ResourceAttributes.CLOUD_PROVIDER, ResourceAttributes.CloudProviderValues.GCP);
```
dashpole, line 47: If we used the GAEResource processor along with other processors, would cloud provider = gcp be added multiple times?

detectors/resources/src/main/java/com/google/cloud/opentelemetry/detectors/GAEResource.java
```diff
@@ -0,0 +1,92 @@
...
+   * Selects the correct method to extract the region, depending on the GAE environment.
+   *
+   * @param attributesBuilder The {@link AttributesBuilder} object to which the extracted region
+   *     would be added.
+   */
+  private void updateAttributesWithRegion(AttributesBuilder attributesBuilder) {
+    if (envVars.get(GAE_ENVIRONMENT) != null && envVars.get(GAE_ENVIRONMENT).equals("standard")) {
+      AttributesExtractorUtil.addCloudRegionFromMetadataUsingRegion(attributesBuilder, metadata);
+    } else {
+      AttributesExtractorUtil.addCloudRegionFromMetadataUsingZone(attributesBuilder, metadata);
```
dashpole, line 89: Should this add availability_zone as well, since we have that information?

detectors/resources/src/main/java/com/google/cloud/opentelemetry/detectors/GAEResource.java
psx95: No, as per the documentation, `AttributesBuilder` a key value store so worst case it would overwrite the cloud provider field.

detectors/resources/src/main/java/com/google/cloud/opentelemetry/detectors/GAEResource.java
psx95: We can add them, but as per the spec document, `availability_zone` is not required. 

No prior resources detections that were added (Cloud Functions, Cloud Run & App Engine) included any additional attributes other than the ones mentioned in the spec.

dashpole: this is fine for now

dashpole (approved):

detectors/resources/src/main/java/com/google/cloud/opentelemetry/detectors/GAEResource.java
psx95: Cool, I think this will be an easy change to make in the future, but it would be better if we first prioritize resolving https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/issues/174 before making changes too the  attributes.

PR #182, opened 2022-11-29
Remove repo level allstar configs

psx95:
The configurations work, but they may turn out to be duplicates since they need to be enabled at the org level. 

Due to this configuration, [these issues](https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/issues?q=is%3Aissue+label%3Aallstar+) were auto-generated. 

Removing these configs from the repo level for now. The auto generated issues will be closed after this PR is merged.

### DIFF
--- a/.allstar/allstar.yaml
+++ b/.allstar/allstar.yaml
@@ -1,2 +1,0 @@
-optConfig:
-  optIn: true
--- a/.allstar/binary_artifacts.yaml
+++ b/.allstar/binary_artifacts.yaml
@@ -1,3 +1,0 @@
-optConfig:
-  optIn: true
-action: issue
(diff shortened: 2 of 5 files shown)


psx95:
> wasn't the point of adding these so that the issues would get opened?

Yes, but after looking further into what was required, I believe the action item was supposed to be taken for org level owners.  Basically, the configuration that we had here should be done on the org level [.github](https://github.com/GoogleCloudPlatform/.github) and [.allstar](https://github.com/GoogleCloudPlatform/.allstar) **repos**.

Only case when we need to add these **directories** at a repo level is when we wish to override some configuration set by the org which is not suited for our use-case.

PR #189, opened 2023-01-26
Add prefix option to metrics exporter

### LINKED ISSUE #186: Allow overriding the default metric prefix
https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/issues/155#issuecomment-1350274398

The default is workload.googleapis.com, which is the successor to custom.googleapis.com.  However, users may want to revert back to old behavior, or write to a different domain entirely.  We should support customizing the metric prefix the exporter uses.  We could do this with a function (e.g. [golang](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/main/exporter/metric/option.go#L91)), or a prefix string (e.g. [nodejs](https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/blob/main/packages/opentelemetry-cloud-monitoring-exporter/src/external-types.ts#L44)

dashpole:
Fixes #186

### DIFF
--- a/exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/MetricConfiguration.java
+++ b/exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/MetricConfiguration.java
@@ -62,6 +62,16 @@ public abstract class MetricConfiguration {
   public abstract String getProjectId();
 
   /**
+   * Returns the prefix prepended to metric names. See
+   * https://cloud.google.com/monitoring/custom-metrics#identifier for more details.
+   *
+   * <p>Defaults to workload.googleapis.com.
+   *
+   * @return the prefix to attach to metrics.
+   */
+  public abstract String getPrefix();
+
+  /**
    * Returns the deadline for exporting to Cloud Monitoring backend.
    *
    * <p>Default value is {{@link MetricConfiguration#DEFAULT_DEADLINE}.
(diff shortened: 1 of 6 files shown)


exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/MetricConfiguration.java
```diff
@@ -61,6 +61,16 @@ public abstract class MetricConfiguration {
    */
   public abstract String getProjectId();
 
+  /**
+   * Returns the prefix prepended to metric names. See
+   * https://cloud.google.com/monitoring/custom-metrics#identifier for more details.
```
psx95, line 66:
```java
/**
   * Returns the prefix prepended to metric names. 
   * @see <a href="https://cloud.google.com/monitoring/custom-metrics#identifier">Custom Metrics Identifiers</a>
   *
   * <p>Defaults to workload.googleapis.com.
   *
   * @return the prefix to attach to metrics.
   */
```

exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/MetricTranslator.java
```diff
@@ -48,7 +49,6 @@ public final class MetricTranslator {
 
   private static final Logger logger = LoggerFactory.getLogger(MetricTranslator.class);
 
-  static final String DESCRIPTOR_TYPE_URL = "workload.googleapis.com/";
```
psx95, line 51: Instead of removing this, we could move this to `MetricConfiguration.java` since this is used in multiple places in the code and has a special meaning (default prefix).

psx95 (approved):

Issue #188, opened 2023-01-19
Create a new release for opentelemetry-operations-java


psx95:
Need to create a new release for this repo which includes the more GCP environments supported for resource detection and upstream OpenTelemetry dependency updates. 

~This would also solve issues like - #141 .~

This would solve #184

dashpole:
This is good to do.

dashpole:
Please compare the MR mapping for java to go before this release

gabroo:
When is this release expected to be pushed? Can you include #189 in it?

psx95:
#189 will be included in the upcoming release. There is no specific date since its P2, but I would expect it in the next 2 weeks.

PR #190, opened 2023-02-03
Update dependencies to remove OSS vulnerabilities

### LINKED ISSUE #191: Update dependencies to remove OSS vulnerabilities
The following dependencies have either known OSS vulnerabilities or they depend on other artifacts with known vulnerabilities. 

These need to be bumped to versions that fix these vulnerabilities. 
 - com.google.cloud:google-cloud-core:2.0.5
 - com.google.cloud:google-cloud-trace:2.0.0
 - com.google.api.grpc:grpc-google-cloud-trace-v2:2.0.0
 - com.google.auto.service:auto-service:1.7.4
 - com.google.auto.service:auto-service-annotations:1.7.4

psx95:
Fixes #191 

Bump dependencies to fix the reported vulnerabilities. 

This needs to be done before addressing #188

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -103,10 +103,10 @@ subprojects {
 	ext {
 		assertjVersion = '3.22.0'
 		autoServiceVersion = '1.0-rc7'
-		autoValueVersion = '1.7.4'
+		autoValueVersion = '1.10.1'
 		slf4jVersion = '1.7.30'
-		googleCloudVersion = '2.0.5'
-		googleTraceVersion = '2.0.0'
+		googleCloudVersion = '2.8.28'
+		googleTraceVersion = '2.1.12'
 		cloudMonitoringVersion = '3.0.0'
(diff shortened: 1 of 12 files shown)


dashpole:
Can you try updating dependencies in individual (or fewer) PRs?  At least update the test image in a different PR from other dependency updates.

psx95:
> Can you try updating dependencies in individual (or fewer) PRs? At least update the test image in a different PR from other dependency updates.

I think the three dependencies being updated share some of the underlying transitive dependencies that bring in the vulnerabilities, so shouldn't they be updated together ? 

I can certainly update the test image in a different PR, but if the underlying concern is too many changes together are breaking the tests, then I think it should be fine since the test failures were caused because of a missing attribute recently added in v0.16.0. 

Let me know your thoughts on this.

dashpole (approved):

Issue #141, opened 2022-02-23
creating TraceExporter disables OpenTelemetry if opentelemetry-opencensus-shim is in use


barend:
I am working on an application that uses OpenTelemetry natively. It also uses the BigTable client, and I'm looking at the opentelemetry-opencensus-shim to include its tracing information.

I ran into an initialisation-order problem while integrating the GCP Trace Exporter for OpenTelemetry. A workaround exists.

### Classpath

These are simply the up-to-date versions of the libraries involved. The one thing worth noting here is that this forces a newer version of the opencensus-api than requested by the GCP SDK packages. This is required to enable the opentelemetry-opencensus-shim.

- com.gooogle.cloud.opentelemetry:exporter-trace:0.20.0
- io.opencensus:opencensus-api version 0.31.0 (any version greater than 0.28.0)
- io.opentelemetry:opentelemetry-sdk version:1.11.0
- io.opentelemetry:opentelemetry-opencensus-shim:1.11.0-alpha

### Problem Description

If the opentelemetry-opencensus-shim is on the classpath, and the `GOOGLE_CLOUD_PROJECT` variable is not set, then constructing the `TraceConfiguration` initializes the `GlobalOpenTelemetry` singleton to `OpenTelemetry.noop()`. This causes the intended configuration of the OpenTelemetry SDK to fail with an exception. If this exception is uncaught, the application fails to start up; if it is caught, it'll work but the telemetry will be inert.

### Problem Details

This happens because of the project ID lookup in `com.google.cloud.ServiceOptions#getDefaultProjectId`, triggered by [`TraceConfiguration#DEFAULT_PROJECT_ID`][ref1]. This function early-returns if the environment variable is present. If it isn't, it initiates a http call to AppEngine, which is instrumented by OpenCensus. This causes initialization of the opencensus-shim ([source][ref2]), which in turn initializes OpenTelemetry. In absence of a configured OpenTelemetry SDK (which we're in the process of configuring), this defaults to the no-op SDK. When we finish building our desired OpenTelemetry config, the `buildAndRegisterGlobal()` call throws an IllegalStateException.

### Sample code

This is an extract of the code I use at startup.

```scala
// The <clinit> of TraceConfiguration ultimately leads to a call to GlobalOpenTelemetry.set()
val gcpTraceExporter = TraceExporter.createWithConfiguration(TraceConfiguration.builder()
    .setFixedAttributes(/* ... */)
    .build())

[...]

PR #193, opened 2023-02-15
Fix trace exporter

psx95:
PR to attempt fixes for #141. 

-----

These changes prevent a crash and tests verified that traces are being generated. This is an early PR for any feedback.

Startup code for testing - 

```java
private static void initializeOTel() {
		traceExporter = TraceExporter.createWithConfiguration(TraceConfiguration.builder()
				.setProjectId(PROJECT_ID)); // contained a live project

		sdkTracerProvider = SdkTracerProvider.builder()
				.setResource(resource)
				.addSpanProcessor(BatchSpanProcessor.builder(traceExporter).build())
				.build();

		openTelemetry = OpenTelemetrySdk.builder()
				.setTracerProvider(sdkTracerProvider)
				.buildAndRegisterGlobal();

		System.out.println("OpenTelemetry initialized Successfully !!!");
}
```

### User Facing Changes (NONE)
The way users interact with this API has not changed, but this PR changes the internal implementation of how a `TraceExporter` object is initialized. The initialization is now deferred till the very first `export` is called.  

#### TraceExporter
1. The original `TraceExporter` class has been converted to `InternalTraceExporter` - this class contains all the logic required and core logic for exporting traces is unchanged.
2. The current `TraceExporter` is a wrapper around the `InternalTraceExporter` and delegates call to `InternalTraceExporter` at the appropriate time. 
3. `TraceExporter` holds a reference to `InternalTraceExporter` which gets initialized when the very first `export` is called. This initialization is only done once in a thread-safe manner (via memoization). 

#### TraceConfiguration & its interaction with TraceExporter
1. User may **not** provide explicit null or empty values for `project ID`. 
2. User provided value for the `project ID` is internally used in a `Supplier`. If the user does not call `setProjectId` to provide a Project ID, the generated `Supplier` will provide default Project ID via `ServiceOptions#getDefaultProjectId`. The supplier used is memoized so the value is retrieved only once. 
3. Building the `TraceConfiguration` object using the builder no longer attempts to fetch the *defaultProjectId* **immediately** via `ServiceOptions.getDefaultProjectId()` call. 
4. Building the `TraceConfiguration` object using builder will now generate a `Supplier` which will resolve to a project ID later when `get()` is called on it. This is done by `InternalTraceExporter` when it is lazy loaded.
5. From the information in (4), this can be inferred that `TraceConfiguration.getProjectId()` will always return whatever user has specified as project ID. 
6. In case the call made to `ServiceOptions.getDefaultProjectId()` (for retrieving default project ID) fails, the `TraceExporter` implementation will be changed to a noop implementation and the user will be notified of this via logs.


exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/DeferredTraceExporterTest.java
```diff
@@ -0,0 +1,35 @@
...
+                .addSpanProcessor(BatchSpanProcessor.builder(gcpTraceExporter).build())
+                .build();
+
+
+        OpenTelemetrySdk otel = OpenTelemetrySdk.builder()
+                .setTracerProvider(tracerProvider)
+                .buildAndRegisterGlobal();
+
+    }
+}
```
dashpole, line 35: nit: newline

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceExporter.java
```diff
@@ -15,134 +15,72 @@
...
-  public CompletableResultCode shutdown() {
-    this.cloudTraceClient.shutdown();
-    return CompletableResultCode.ofSuccess();
-  }
+    @Override
+    public CompletableResultCode shutdown() {
+        if (internalTraceExporter != null) {
+            return internalTraceExporter.shutdown();
+        } else {
+            return CompletableResultCode.ofFailure();
```
jsuereth, line 83: Is it actually a failure ot attempt to shutdown something that never started?

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceExporter.java
```diff
@@ -15,134 +15,72 @@
...
+        return internalTraceExporter.flush();
     }
 
-    cloudTraceClient.batchWriteSpans(projectName, spans);
-    return CompletableResultCode.ofSuccess();
-  }
+    @Override
+    public CompletableResultCode export(@Nonnull Collection<SpanData> spanDataList) {
+        try {
+            if (internalTraceExporter == null) {
```
jsuereth, line 68: FYI - this type of delegation is not threadsafe. 


I suggest you use an `AtomicReference<SpanExporter>`, and you can do something like:

```
while (internalTraceExporter.get() == null) {
  internalTraceExporter.compareAndSet(null, createActualTraceExporter());
}
```

This will ensure that trace exporter is created (Possibly more than once, but only one is kept), and the AtomicReference ensures you'll be threadsafe.

You can also look up "typesafe singleton pattern in Java" for other options.

jsuereth (changes requested):

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceExporter.java
psx95: So this is the case where somehow `shutdown` is called before call to an `export` which initializes the `internalTraceExporter`. 

So it's something that never started. Should we send some other code instead of Failure ?

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceExporter.java
psx95: Yup, that's a good point. I will make the change.

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceExporter.java
```diff
@@ -15,134 +15,99 @@
...
-  public CompletableResultCode export(Collection<SpanData> spanDataList) {
-    List<Span> spans = new ArrayList<>(spanDataList.size());
-    for (SpanData spanData : spanDataList) {
-      spans.add(translator.generateSpan(spanData, projectId));
+  public CompletableResultCode export(@Nonnull Collection<SpanData> spanDataList) {
+    try {
+      internalTraceExporter.compareAndSet(null, createActualTraceExporter());
+      return internalTraceExporter.get().export(spanDataList);
+    } catch (IOException e) {
+      e.printStackTrace();
```
psx95, line 100: Remove the stacktrace

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceExporter.java
```diff
@@ -15,134 +15,115 @@
...
-import com.google.cloud.trace.v2.TraceServiceSettings;
-import com.google.cloud.trace.v2.stub.TraceServiceStub;
-import com.google.common.collect.ImmutableMap;
-import com.google.devtools.cloudtrace.v2.AttributeValue;
-import com.google.devtools.cloudtrace.v2.ProjectName;
-import com.google.devtools.cloudtrace.v2.Span;
-import io.grpc.ManagedChannelBuilder;
+import com.google.cloud.ServiceOptions;
 import io.opentelemetry.sdk.common.CompletableResultCode;
+import io.opentelemetry.sdk.internal.ThrottlingLogger;
```
punya, line 20: The package name suggests we aren't supposed to depend on this class.

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceExporter.java
```diff
@@ -15,134 +15,115 @@
...
 
-  public static SpanExporter createWithConfiguration(TraceConfiguration configuration)
-      throws IOException {
-    String projectId = configuration.getProjectId();
-    TraceServiceStub stub = configuration.getTraceServiceStub();
-
-    // TODO: Remove stub.
-    if (stub == null) {
-      TraceServiceSettings.Builder builder = TraceServiceSettings.newBuilder();
+  private static SpanExporter generateStubTraceExporter(TraceConfiguration.Builder configBuilder) {
```
punya, line 45: Are you using "stub" to mean something like "placeholder" or "temporary" here? If so, consider a different name because "stub" is generally used to mean [RPC stub](https://en.wikipedia.org/wiki/Stub_(distributed_computing)).

build.gradle
```diff
@@ -149,6 +150,7 @@ subprojects {
 			spring_boot_starter_web          : "org.springframework.boot:spring-boot-starter-web:${springWebVersion}",
 			spring_cloud_starter_openfeign   : "org.springframework.cloud:spring-cloud-starter-openfeign:${springOpenFeignVersion}",
 			spring_cloud_sleuth_otel_autoconf: "org.springframework.cloud:spring-cloud-sleuth-otel-autoconfigure:${springOtelVersion}",
+			opencensus_shim: "io.opentelemetry:opentelemetry-opencensus-shim:${opencensusShim}-alpha"
```
punya, line 153: After this change, our tests always (rather than never) have the OC shim on the classpath. Can we test both scenarios, e.g. by having a subproject for the sole purpose of testing OC interactions?

@dashpole curious if you think OC shim interactions should be tested here vs in the OC shim.

exporters/trace/build.gradle
```diff
@@ -13,6 +13,9 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+plugins {
```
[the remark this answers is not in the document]
psx95: It is not needed, but this just generates a fat jar for exporter-trace dependency when we run `./gradlew shadowJar` task for the project so that exporter-trace can be consumed as an independent dependency. 

It is just for convenience.

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/InternalTraceExporter.java
```diff
@@ -0,0 +1,174 @@
+/*
+ * Copyright 2022 Google
```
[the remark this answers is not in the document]
psx95: This part is autogenerated by the task `./gradlew spotlessApply`, if we manually change it to 2023, the stylecheck in GitHub action fails. 
I have a tracking bug - #196 that should address this issue. It's an easy fix, but would touch a lot of files, so I did not do it here.

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/InternalTraceExporter.java
```diff
@@ -0,0 +1,174 @@
...
+  private final CloudTraceClient cloudTraceClient;
+  private final ProjectName projectName;
+  private final String projectId;
+  private final TraceTranslator translator;
+
+  private static final Map<String, String> HEADERS =
+      Map.of("User-Agent", "opentelemetry-operations-java/" + TraceVersions.EXPORTER_VERSION);
+  private static final HeaderProvider HEADER_PROVIDER = () -> HEADERS;
+
+  private static InternalTraceExporter createWithClient(
```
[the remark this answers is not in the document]
psx95: So one thing to note here is that this is the exact same class as the `TraceExporter` in the current [main](https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/blob/main/exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceExporter.java). 

It has just been renamed to `InternalTraceExporter`. In order to minimize the changes, I did not edit any existing logic.

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/InternalTraceExporter.java
```diff
@@ -0,0 +1,174 @@
...
+      Map<String, AttributeValue> fixedAttributes) {
+    return new InternalTraceExporter(
+        projectId, cloudTraceClient, attributeMappings, fixedAttributes);
+  }
+
+  static SpanExporter createWithConfiguration(TraceConfiguration configuration) throws IOException {
+    String projectId = configuration.getProjectId();
+    TraceServiceStub stub = configuration.getTraceServiceStub();
+
+    // TODO: Remove stub.
```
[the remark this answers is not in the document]
psx95: As mentioned this is the current existing `TraceExporter` class from the main. This TODO has been there for some while now, but I have not looked into why this is needed. But I agree that it should be handled in a separate bug, since it warrants some investigation.

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/InternalTraceExporter.java
```diff
@@ -0,0 +1,174 @@
...
+          configuration.getFixedAttributes());
+    }
+    return InternalTraceExporter.createWithClient(
+        projectId,
+        new CloudTraceClientImpl(TraceServiceClient.create(stub)),
+        configuration.getAttributeMapping(),
+        configuration.getFixedAttributes());
+  }
+
+  /**
```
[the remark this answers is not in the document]
psx95: Yes, this would be better.

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/InternalTraceExporter.java
```diff
@@ -0,0 +1,174 @@
...
+
+  InternalTraceExporter(
+      String projectId,
+      CloudTraceClient cloudTraceClient,
+      ImmutableMap<String, String> attributeMappings,
+      Map<String, AttributeValue> fixedAttributes) {
+    this.projectId = projectId;
+    this.cloudTraceClient = cloudTraceClient;
+    this.projectName = ProjectName.of(projectId);
+    this.translator = new TraceTranslator(attributeMappings, fixedAttributes);
```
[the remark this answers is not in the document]
psx95: That does sound better, I will open a separate issue that would look into in the existing TODOs of this class and do the necessary refactoring, this would help in keeping the scope of this PR small.

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceExporter.java
psx95: Yes, I was indeed using it for that. I can change the naming then.

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/InternalTraceExporter.java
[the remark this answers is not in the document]
psx95: I will see if I can manually fix other style issues and then change header so that GitHub actions does not fail. Should be possible though.

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/InternalTraceExporter.java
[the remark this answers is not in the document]
psx95: Created issue #198 to track this and linked the issue number in the code.

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceExporter.java
psx95: For lack of a better name, replaced it with `generateSkeletonTraceExporter`, other alternative considered was `generateTraceExporterTemplate`, but it seemed to me it was not clarifying the semantics well enough. But I am open to suggestions on this.

build.gradle
psx95: IMO, the OC shim interactions should be tested here since we are making the changes here to fix the issue.

The idea of having a separate project to test the *without shim* scenario would definitely add some value, but it may not be necessary since without OC shim might be the *general case* (I may be wrong here). Do we have a reason to believe that removing the dependency from OC shim would cause any issues ?

psx95: I have removed this dependency and the unit test that attempted to test the failure case with this dependency. When running as GitHub actions, the tests were behaving differently from their local run. 

Let me know if having such a unit test in exporter-trace is still a good idea. Punya's suggestion of having a subproject which specifies a dependency on both - shim & exporter trace seems like a good place for it, but not sure if the extra effort is worth the gains.

build.gradle
```diff
@@ -120,6 +120,7 @@ subprojects {
 		springOtelVersion = '1.0.0-M8'
 		cloudEventsCoreVersion = '2.4.0'
 		cloudFunctionsFrameworkApiVersion = '1.0.4'
+		opencensusShim = '1.23.1'
```
psx95, line 123: Need to remove this, if OC shim is not going to be tested here.

build.gradle
psx95: This is moot now, since deferred approach is the only approach that is currently available. So we cannot replicate the failure situation now. 

Current unit tests verify that the trace exporter works without any issues **even if** OC-shim is on the classpath.

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceConfiguration.java
```diff
@@ -231,17 +240,44 @@ public final Builder addAttributeMapping(String otelKey, String gcpKey) {
...
+      TraceConfiguration builtTraceConfiguration = autoBuild();
+      builtTraceConfiguration.setProjectIdProvider(projectIdProvider);
+      return builtTraceConfiguration;
+    }
+
+    @Nonnull
+    private Supplier<String> generateProjectIdProvider() {
+      String userProvidedProjectId = getUserProvidedProjectId();
+      if (Strings.isNullOrEmpty(userProvidedProjectId)) {
+        return () -> {
```
psx95, line 260: Should this be memoized ?

Currently this is only being called once.

Since the Supplier generated can only be retrieved by a package-private getter, it is not exposed to the end user and we have control on how often this used. 
`get()` on this supplier is called during the initialization of trace exporter. So per `TraceExporter` it will be called once. 

Thoughts @punya @jsuereth ?

build.gradle
punya, line 123: For consistency, this variable should be called `opencensusShimVersion`. Also, the value should include `-alpha`.

exporters/trace/build.gradle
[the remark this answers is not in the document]
punya: Please remove this change because it's unrelated to this PR.

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/NoopTraceExporter.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.Objects;
+import javax.annotation.Nonnull;
+import javax.annotation.concurrent.ThreadSafe;
+
+/**
+ * A noop implementation of a {@link SpanExporter}. This class is implemented as a singleton and
+ * return a successful result on all public methods without performing any action.
+ */
+@ThreadSafe
+public class NoopTraceExporter implements SpanExporter {
```
punya, line 31: Our primary class is called `TraceExporter` because it exports to the `Cloud Trace` product. In this case, `NoopSpanExporter` is a more appropriate name.

Also, is there any reason to make this public rather than package-private?

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceExporter.java
```diff
@@ -15,134 +15,108 @@
...
+      synchronized (this) {
+        try {
+          internalTraceExporter.compareAndSet(null, createActualTraceExporter());
+        } catch (IOException e) {
+          // unable to create actual trace exporter
+          logger.log(
+              Level.WARNING,
+              String.format("Unable to initialize trace exporter. Error %s", e.getMessage()));
+          logger.log(Level.INFO, "Setting TraceExporter to noop.");
+          internalTraceExporter.set(InternalTraceExporter.noop());
```
punya, line 107:
```suggestion
          internalTraceExporter.set(new NoopSpanExporter());
```

IMO, there's no real benefit to having the no-op exporter be a singleton. This would let us get rid of the singleton code as well as the indirection of `SpanExporter.noop()`.

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceConfiguration.java
```diff
@@ -60,6 +59,16 @@ public abstract class TraceConfiguration {
           .put("thread.id", "/tid")
           .build();
 
+  private Supplier<String> projectIdProvider;
```
punya, line 62: * An auto-value class should not have concrete fields or setters.
* This should be named `projectIdSupplier` to be consistent with the type name.

One way to do this would be
1. Remove the field
2. Add an abstract method `getProjectIdSupplier` to `TraceConfiguration`
3. Change the existing abstract method `getProjectId` to a concrete, [memoized](https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/pull/193/files?diff=split&w=1) method that calls `getProjectIdSupplier().get()`.
4. Add a concrete method `setProjectId` to `TraceConfiguration.Bulder` that calls `setProjectIdSupplier(() -> id)`.

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceConfiguration.java
```diff
@@ -127,7 +136,7 @@ public abstract class TraceConfiguration {
    */
   public static Builder builder() {
     return new AutoValue_TraceConfiguration.Builder()
-        .setProjectId(DEFAULT_PROJECT_ID)
+        .setProjectId("")
```
punya, line 139: If you follow my suggestion earlier, this should become `.setProjectIdSupplier(ServiceOptions::getDefaultProjectId)`

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceConfiguration.java
punya: If you follow my suggestions from the other comment, you'll end up memoizing but in a slightly different place.

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceExporter.java
```diff
@@ -15,134 +15,108 @@
...
 import io.opentelemetry.sdk.trace.data.SpanData;
 import io.opentelemetry.sdk.trace.export.SpanExporter;
 import java.io.IOException;
-import java.util.ArrayList;
 import java.util.Collection;
-import java.util.List;
-import java.util.Map;
+import java.util.Objects;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.logging.Level;
```
punya, line 26: We should probably use SLF4J rather than JUL. Thoughts @jsuereth ?

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceExporter.java
```diff
@@ -15,134 +15,108 @@
...
-        Credentials credentials =
-            configuration.getCredentials() == null
-                ? GoogleCredentials.getApplicationDefault()
-                : configuration.getCredentials();
-        builder.setCredentialsProvider(
-            FixedCredentialsProvider.create(checkNotNull(credentials, "credentials")));
-        builder.setEndpoint(configuration.getTraceServiceEndpoint());
-        builder.setHeaderProvider(HEADER_PROVIDER);
-      }
+  private static SpanExporter generateSkeletonTraceExporter(TraceConfiguration config) {
```
punya, line 44: Consider inlining this method.

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceExporter.java
```diff
@@ -15,134 +15,108 @@
...
-          new CloudTraceClientImpl(TraceServiceClient.create(builder.build())),
-          configuration.getAttributeMapping(),
-          configuration.getFixedAttributes());
-    }
-    return TraceExporter.createWithClient(
-        projectId,
-        new CloudTraceClientImpl(TraceServiceClient.create(stub)),
-        configuration.getAttributeMapping(),
-        configuration.getFixedAttributes());
+  private SpanExporter createActualTraceExporter() throws IOException {
```
punya, line 48: Consider inlining this method.

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceExporter.java
```diff
@@ -15,134 +15,108 @@
...
+   */
+  public static SpanExporter createWithConfiguration(TraceConfiguration configuration) {
+    return generateSkeletonTraceExporter(configuration);
   }
 
   @Override
   public CompletableResultCode flush() {
-    // We do no exporter buffering of spans, so we're always flushed.
-    return CompletableResultCode.ofSuccess();
+    if (internalTraceExporter.get() == null) {
```
punya, line 89: Can we unconditionally return success for this method, without even checking whether the field has been set?

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceExporter.java
```diff
@@ -15,134 +15,108 @@
...
   }
 
   @Override
-  public CompletableResultCode export(Collection<SpanData> spanDataList) {
-    List<Span> spans = new ArrayList<>(spanDataList.size());
-    for (SpanData spanData : spanDataList) {
-      spans.add(translator.generateSpan(spanData, projectId));
+  public CompletableResultCode export(@Nonnull Collection<SpanData> spanDataList) {
+    if (Objects.isNull(internalTraceExporter.get())) {
+      synchronized (this) {
```
punya, line 98: I'm confused by the synchronization model here - usually an `AtomicReference` is used to avoid synchronization. You might be able to avoid a lot of this complexity by changing `internalTraceExporter` to `Suppliers.memoized(() -> new InternalTraceExporter(this.config))` and making it do all the work.

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/InternalTraceExporter.java
```diff
@@ -0,0 +1,158 @@
...
+  private static InternalTraceExporter createWithClient(
+      String projectId,
+      CloudTraceClient cloudTraceClient,
+      ImmutableMap<String, String> attributeMappings,
+      Map<String, AttributeValue> fixedAttributes) {
+    return new InternalTraceExporter(
+        projectId, cloudTraceClient, attributeMappings, fixedAttributes);
+  }
+
+  static SpanExporter createWithConfiguration(TraceConfiguration configuration) throws IOException {
```
punya, line 69: Consider wrapping the IOException at this layer and removing it from the signature. Since our external API no longer has an IOException, this just adds noise to our implementation.

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/NoopTraceExporter.java
psx95: No, I will change it to package private, it was an oversight.

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceExporter.java
psx95: My major motivation to do this was the fact that all objects would be exactly same for noop implementation, so in case of multiple TraceExporters, this could be helpful. 

But I can see the `new` call here would be quite inexpensive so I don't have a strong opinion. I can remove singleton implementation here.

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceExporter.java
psx95: Looking at this with fresh eyes, I agree with this. Thanks.

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceExporter.java
psx95: Would defer to whatever is decided here. But is there some benefit here for SLF4J ?

IMO though SLF4J is helpful when there are a lot of logs and they needed some sort of formatting or structure applied to them. This would also require additional config.

For simple logging Java logger would be better.

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceExporter.java
psx95: Technically, we can. But the purpose of this class it defer business logic to the `internalTraceExporter`. 

However, I was thinking if failure is the correct code here. Any thoughts ? This condition would happen if flush is called before export somehow.

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceConfiguration.java
psx95: This would be an API change. I am assuming this would **replace** the current `setProjectId` ? Because exposing both setter would be confusing for the user.

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/InternalTraceExporter.java
psx95: So this exception may be thrown from two possible places, handling it at the `InternalTraceExporter` level makes the implementation a little ugly since both need to be handled. Also, `InternalTraceExporter` is the class that actually has the implementation so handling it there would introduce more noise to our implementation detail. 

I think it makes sense to keep it in the function signature here since it has a package private access and this exception is handled in `TraceExporter` for both cases.

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceConfiguration.java
punya: No I'm expecting this to be an additive change (not a backwards compatibility break). We can resolve the confusion by documenting the provider setter.

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceExporter.java
punya: SLF4J is a logging facade, not a specific logging library. It allows the end-user of the application to decide which library to use (like Log4J, Logback, etc).

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceExporter.java
```diff
@@ -15,134 +15,84 @@
...
+   */
+  public static SpanExporter createWithConfiguration(TraceConfiguration configuration) {
+    return new TraceExporter(configuration);
   }
 
   @Override
   public CompletableResultCode flush() {
-    // We do no exporter buffering of spans, so we're always flushed.
-    return CompletableResultCode.ofSuccess();
+    return internalTraceExporterSupplier.get().flush();
```
psx95, line 86: Having a `NoopSpanExporter` facilitates for a much cleaner and simple looking code, a downside for this is that if for some reason `flush()` or `shutdown()` is called prior to a call to `export()`, the current code would first initialize the `internalTraceExporter` - only to flush or shutdown. 

We could use an `AtomicBoolean` to keep track if its initialized and conditionally skip initializing;  simply returning success codes, but this would make code a less elegant here.

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/NoopSpanExporter.java
```diff
@@ -0,0 +1,53 @@
...
+ */
+class NoopSpanExporter implements SpanExporter {
+
+  NoopSpanExporter() {
+    // prevent explicit public call to default constructor
+  }
+
+  @Override
+  public CompletableResultCode export(@Nonnull Collection<SpanData> spans) {
+    return CompletableResultCode.ofFailure();
```
psx95, line 36: The rationale behind this is that a NoopSpanExporter (a no-op implementation of TraceExporter) can be expected to `flush` or `shutdown`, but is not able to `export` - hence returning failure.

punya (approved):
🎉

exporters/trace/build.gradle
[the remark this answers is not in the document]
jsuereth: Should this have been removed?

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/NoopSpanExporter.java
jsuereth: This should succeed.  The only reason here is if you have hope of recovery or you need the upstream system to "stop".

Unfortunately, given the runtime you're in, there's no good way to stop, you have no hope of recovery and you've already logged the actual error.  This should return `ofSuccess`

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceExporter.java
```diff
@@ -15,134 +15,86 @@
...
-  public static SpanExporter createWithDefaultConfiguration() throws IOException {
-    TraceConfiguration configuration = TraceConfiguration.builder().build();
-    return TraceExporter.createWithConfiguration(configuration);
+  private static final Logger logger = LoggerFactory.getLogger(TraceExporter.class);
+
+  private final Supplier<SpanExporter> internalTraceExporterSupplier;
+
+  private TraceExporter(TraceConfiguration configuration) {
+    this.internalTraceExporterSupplier =
+        Suppliers.memoize(
```
jsuereth, line 38: Nice use of the memoize function.

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceExporter.java
```diff
@@ -15,134 +15,86 @@
...
+            () -> {
+              try {
+                return InternalTraceExporter.createWithConfiguration(configuration);
+              } catch (IOException e) {
+                logger.warn(
+                    "Unable to initialize TraceExporter. Export operation failed, switching to NoopSpanExporter",
```
jsuereth, line 44: Add some more context here if you can.

"Unable to initialize Google Cloud TraceExporter" e.g.

exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/TraceConfigurationTest.java
```diff
@@ -78,29 +85,40 @@ public void setAllConfigurationFields() {
...
   @Test
   public void allowToUseDefaultProjectId() {
-    String defaultProjectId = ServiceOptions.getDefaultProjectId();
-    // some test providers might not have project IDs set up
-    if (defaultProjectId != null) {
-      TraceConfiguration configuration = TraceConfiguration.builder().build();
-
-      assertEquals(defaultProjectId, configuration.getProjectId());
+    // some test providers might not have project IDs set up - so we use mocks
+    try (MockedStatic<ServiceOptions> mockedServiceOptions =
```
jsuereth, line 98: I don't think you need to create the entire `TraceExporter` in this mock.

instead you should be able to just pull the `ProjectId` provider from the default configuration and check that, like the original test.

That would dramatically simplify this test.

exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/TraceExporterTest.java
```diff
@@ -15,25 +15,129 @@
...
+  @Mock private TraceServiceStub mockedTraceServiceStub;
+
+  @After
+  public void tearDown() {
+    GlobalOpenTelemetry.resetForTest();
+  }
+
+  @SuppressWarnings("ResultOfMethodCallIgnored")
   @Test
   public void createWithConfiguration() {
```
jsuereth, line 58: This is no longer just a "will the constructor work" unit test.  You're fully testing to a  more integration-level here.

As such I'd update the test name to denote what it's doing *or* limit what you test to just what was done previously.

exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/TraceExporterTest.java
```diff
@@ -15,25 +15,129 @@
...
 
+      mockedTraceServiceClient.verify(
+          () -> TraceServiceClient.create(Mockito.eq(mockedTraceServiceStub)));
+      Mockito.verify(this.mockedTraceServiceClient)
+          .batchWriteSpans((ProjectName) Mockito.any(), Mockito.anyList());
+    }
+  }
+
+  @Test
+  public void createWithDefaultConfiguration() {
```
jsuereth, line 83: Similar comment as above.

exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/TraceExporterTest.java
```diff
@@ -15,25 +15,129 @@
...
+      mockedTraceServiceClient.verify(
+          Mockito.times(1), () -> TraceServiceClient.create((TraceServiceSettings) Mockito.any()));
+      mockedServiceOptions.verify(Mockito.times(1), ServiceOptions::getDefaultProjectId);
+      Mockito.verify(this.mockedTraceServiceClient)
+          .batchWriteSpans((ProjectName) Mockito.any(), Mockito.anyList());
     }
   }
+
+  @Test
+  public void createNoopSpanExporterIfExceptionThrown() {
```
jsuereth, line 112: Similar comment as above.  You're verifying flush/shutdown/export.

This doesn't really make sure Noop span exporter is created, you're more testing that a failure in construction doesn't cause later failures in export.

exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/TraceConfigurationTest.java
psx95: The original test had a conditional check if `ServiceOptions.getDefaultProjectId()` would return null, the test was essentially no-op. 

This `== null` condition returned true when run on GitHub Actions environment where GCP metadata does not exist.  
Mocking makes this test run independent of environment. 

I can add `null` condition check and make test simple if this seems too complex, but then this test is not adding too much value when run as a GitHub action.

exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/TraceExporterTest.java
psx95: Right, this now verifies that export is working as well. I will rename this to `verifyExporterWorksWithConfiguration` .

exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/TraceExporterTest.java
psx95: Agreed, good point. I will rename this test to reflect the scenario.

Although, just to be clear, a failure in construction here will end up resulting in failure code (for export), but not cause additional exceptions.

exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/TraceConfigurationTest.java
jsuereth: I'm not objecting to Mockito, I'm objecting to how you're getting the call to `getDefaultProjectId` to trigger.

Specifically, you're instantiating the exporter and calling `exporter.export(Collections.emptyList())`.

I think you only have to construct default Configuration and then use the ProjectIdSupplier.

In other words, if this is a unit test, you should try to test only using the TraceConfiguration class and its dependencies.

exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/TraceExporterTest.java
```diff
@@ -15,25 +15,130 @@
...
+          .when(() -> InternalTraceExporter.createWithConfiguration(Mockito.any()))
+          .thenThrow(IOException.class);
+
+      SpanExporter traceExporter = TraceExporter.createWithDefaultConfiguration();
+      assertNotNull(traceExporter);
+
+      SpanData mockSpanData = Mockito.mock(SpanData.class);
+      // verify trace exporter still works without any additional exceptions
+      assertEquals(
+          CompletableResultCode.ofFailure(),
```
jsuereth, line 125: This should not be failing

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/NoopSpanExporter.java
jsuereth: Reopening this.  This should be returning a success status code.  The only thing returning failure here is likely to do would be cause log-spam (at best) or down the SDK completely (at worst), providing no additional value to the user over the initial failure log statement that was dumped.

jsuereth (commented):
The Noop version should not be returning any failures.

exporters/trace/src/test/java/com/google/cloud/opentelemetry/trace/TraceExporterTest.java
psx95: Fixed.

jsuereth (approved):

PR #194, opened 2023-02-15
Replaced concrete type TraceExporter with interface type SpanExporter

psx95:
### Breaking Change
 - Replacing the usages of concrete type `TraceExporter` with interface type `SpanExporter` for public methods in `TraceExporter` class.

This change should help with the future work of adding a Lazy initialized / Deferred trace exporter.


punya (approved):

jsuereth (approved):
LGTM - but we'll need to bump a major version here for users.

jsuereth:
Looks like you need to update the e2e test code for this breaking change.

psx95:
> LGTM - but we'll need to bump a major version here for users.

Yes, I have a draft PR here -  https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/pull/193, will make a release once this is completed as well. 

For major versions, 0.24.0 &rarr; 0.25.0 seems ok ?

psx95:
> Looks like you need to update the e2e test code for this breaking change.

Seemed like an intermittent issue - for some reason GCB decided to skip build-image step since it "already existed" and so it was not re-built for the commit ID in this PR - which is why no tests were able to find that image.

PR #195, opened 2023-02-16
Lazy load project ID in TraceConfiguration

### LINKED ISSUE #141: creating TraceExporter disables OpenTelemetry if opentelemetry-opencensus-shim is in use

psx95:
Fixes #141 

Lazy loading Project ID in trace configuration prevents crashes when using the exporter-trace with opencensus-shim. 

#### Test Setup
 - Simple Hello world Java app running on spring-boot. 
 - Traces generated were visible in Google Cloud Trace

Initialization code
```java
private static void initializeOTel() {
		try {
			traceExporter = TraceExporter.createWithConfiguration(TraceConfiguration.builder()
					.setProjectId(PROJECT_ID).build()); //
		} catch (IOException e) {
			e.printStackTrace();
		}

		sdkTracerProvider = SdkTracerProvider.builder()
				.setResource(resource)
				.addSpanProcessor(BatchSpanProcessor.builder(traceExporter).build())
				.build();

		openTelemetry = OpenTelemetrySdk.builder()
				.setTracerProvider(sdkTracerProvider)
				.buildAndRegisterGlobal();

		System.out.println("OpenTelemetry initialized Successfully !!!");
}
```

### DIFF
--- a/exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceConfiguration.java
+++ b/exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceConfiguration.java
@@ -228,7 +228,13 @@ public abstract class TraceConfiguration {
      */
     public TraceConfiguration build() {
       // If project ID is not set, attempt to get the Default Project ID
-      if (Strings.isNullOrEmpty(getProjectId())) {
+      String currentProjectId;
+      try {
+        currentProjectId = getProjectId();
+      } catch (IllegalStateException e) {
+        currentProjectId = null;
+      }
+      if (Strings.isNullOrEmpty(currentProjectId)) {
         setProjectId(Strings.nullToEmpty(ServiceOptions.getDefaultProjectId()));
       }
       // Make a defensive copy of fixed attributes.
--- a/exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceConfiguration.java
+++ b/exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceConfiguration.java
@@ -25,20 +25,18 @@ import com.google.common.base.Preconditions;
 import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableMap;
 import com.google.devtools.cloudtrace.v2.AttributeValue;
+
+import javax.annotation.Nullable;
+import javax.annotation.concurrent.Immutable;
 import java.time.Duration;
 import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.Map;
-import javax.annotation.Nullable;
-import javax.annotation.concurrent.Immutable;
 
 /** Configurations for {@link TraceExporter}. */
 @AutoValue
 @Immutable
 public abstract class TraceConfiguration {
-
-  private static final String DEFAULT_PROJECT_ID =
-      Strings.nullToEmpty(ServiceOptions.getDefaultProjectId());
 
   @VisibleForTesting static final Duration DEFAULT_DEADLINE = Duration.ofSeconds(10, 0);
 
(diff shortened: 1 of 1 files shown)


punya:
I don't understand how this fixes the problem. If you remove the `.setProjectId(PROJECT_ID)` from your sample code, it seems like we'll still have the following sequence of events

1. Build the configuration -> call metadata server -> touch OC -> implicitly set the OTel global
2. Try to explicitly set the global, which will fail

psx95:
Yes, that's a good point, let me retrace my steps and see if I missed something. 
Thanks.

psx95:
Yup, you are right when project ID is not explicitly set, it leads to the same issue. 
Closing this PR since #193 contains this change and additional changes required to make it work for the case when Project ID is not set.

PR #206, opened 2023-03-06
Fix exporter-trace example & add README

### LINKED ISSUE #205: Fix exporter-trace example
The current exporter-trace example (under examples directory) crashes on running. 

Address the TODOs and fix the example.

psx95:
Fixes #205 .

### DIFF
--- a/examples/trace/src/main/java/com/google/cloud/opentelemetry/example/trace/TraceExporterExample.java
+++ b/examples/trace/src/main/java/com/google/cloud/opentelemetry/example/trace/TraceExporterExample.java
@@ -29,7 +29,7 @@ import java.util.Random;
 import java.util.concurrent.TimeUnit;
 
 public class TraceExporterExample {
-  private static final String INSTRUMENTATION_SCOPE_NAME = TraceExporter.class.getName();
+  private static final String INSTRUMENTATION_SCOPE_NAME = TraceExporterExample.class.getName();
   private static final Random random = new Random();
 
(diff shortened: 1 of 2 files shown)


examples/trace/src/main/java/com/google/cloud/opentelemetry/example/trace/TraceExporterExample.java
```diff
@@ -17,23 +17,23 @@
...
 import io.opentelemetry.sdk.trace.export.SpanExporter;
 import java.time.Duration;
 import java.util.Random;
+import java.util.concurrent.TimeUnit;
 
 public class TraceExporterExample {
-  // TODO: Update this to NON-global OpenTelemetry.  Only APIs/Frameworks should be using global.
-  private static final Tracer tracer =
-      GlobalOpenTelemetry.getTracer("io.opentelemetry.example.TraceExporterExample");
+  private static final String INSTRUMENTATION_SCOPE_NAME = TraceExporter.class.getName();
```
jsuereth, line 32: NIt: This should be the TraceExporterExample, to match what was there previously.

jsuereth (approved):
One nit, but pretty important for an example.  We don't want folks attachign trace instrumentation against our Exporter class.

psx95:
> One nit, but pretty important for an example. We don't want folks attachign trace instrumentation against our Exporter class.

Agreed, I updated it.

PR #207, opened 2023-03-08
Fix metrics exporter integration

### LINKED ISSUE #202: creating MetricExporter disables OpenTelemetry if opentelemetry-opencensus-shim is in use
Very similar to  #141 and should be resolved in a similar way. Look at #141 for details.

### Summary: 
Using exporter-trace dependency with opencensus-shim causes a failure. The issue was able to be reproduced with the same error - 

```
Exception in thread "main" java.lang.IllegalStateException: GlobalOpenTelemetry.set has already been called. GlobalOpenTelemetry.set must be called only once before any calls to GlobalOpenTelemetry.get. If you are using the OpenTelemetrySdk, use OpenTelemetrySdkBuilder.buildAndRegisterGlobal instead. Previous invocation set to cause of this exception.
	at io.opentelemetry.api.GlobalOpenTelemetry.set(GlobalOpenTelemetry.java:99)
	at io.opentelemetry.sdk.OpenTelemetrySdkBuilder.buildAndRegisterGlobal(OpenTelemetrySdkBuilder.java:84)
	at com.example.springboot.Application.initializeOTel(Application.java:96)
	at com.example.springboot.Application.main(Application.java:44)
Caused by: java.lang.Throwable
	at io.opentelemetry.api.GlobalOpenTelemetry.set(GlobalOpenTelemetry.java:107)
	at io.opentelemetry.api.GlobalOpenTelemetry.get(GlobalOpenTelemetry.java:77)
	at io.opentelemetry.api.GlobalOpenTelemetry.getTracer(GlobalOpenTelemetry.java:126)
	at io.opentelemetry.opencensusshim.OpenTelemetrySpanBuilderImpl.<clinit>(OpenTelemetrySpanBuilderImpl.java:52)
	at io.opentelemetry.opencensusshim.OpenTelemetryTracerImpl.spanBuilderWithExplicitParent(OpenTelemetryTracerImpl.java:42)
	at io.opencensus.trace.Tracer.spanBuilder(Tracer.java:308)
	at com.google.api.client.http.HttpRequest.execute(HttpRequest.java:865)
	at com.google.cloud.ServiceOptions.getAppEngineProjectIdFromMetadataServer(ServiceOptions.java:503)
	at com.google.cloud.ServiceOptions.getAppEngineProjectId(ServiceOptions.java:472)
	at com.google.cloud.ServiceOptions.getDefaultProjectId(ServiceOptions.java:379)
	at com.google.cloud.opentelemetry.metric.MetricConfiguration.<clinit>(MetricConfiguration.java:42)
	at com.google.cloud.opentelemetry.metric.GoogleCloudMetricExporter.createWithDefaultConfiguration(GoogleCloudMetricExporter.java:76)
	at com.example.springboot.Application.initializeOTel(Application.java:79)
	... 1 more

```

psx95:
Fixes #202 

Also, updates the existing example to replace the infinite loop with fixed number of iterations. 

This PR implements a deferred exporter approach, similar to #193. 

### Breaking Changes
 - `GoogleCloudMetricExporter.createWithDefaultConfiguration` now returns an interface type: `GoogleCloudMetricExporter` &rarr; `MetricExporter` 
 - `GoogleCloudMetricExporter.createWithConfiguration` now returns an interface type: `GoogleCloudMetricExporter` &rarr; `MetricExporter` 

### User Facing API changes 
 - NONE
*Users interaction with this library's API will not be affected (other than the breaking change)*

#### GoogleCloudMetricExporter
1. The original `GoogleCloudMetricExporter` class has been converted to `InternalMetricExporter` - this class contains all the logic required and core logic for exporting metrics is unchanged.
2. The current `GoogleCloudMetricExporter` is a wrapper around the `InternalMetricExporter` and delegates call to `InternalMetricExporter` at the appropriate time. 
3. `GoogleCloudMetricExporter` holds a reference to `InternalTraceExporter` which gets initialized when the very first `export` is called. This initialization is only done once in a thread-safe manner (via memoization). 

#### MetricConfiguration & its interaction with GoogleCloudMetricExporter
1. User may **not** provide explicit null or empty values for `project ID`. 
2. User provided value for the `project ID` is internally used in a `Supplier`. If the user does not call `setProjectId` to provide a Project ID, the generated `Supplier` will provide default Project ID via `ServiceOptions#getDefaultProjectId`. The supplier used is memoized so the value is retrieved only once. 
3. Building the `MetricConfiguration` object using the builder no longer attempts to fetch the *defaultProjectId* **immediately** via `ServiceOptions.getDefaultProjectId()` call. 
4. Building the `MetricConfiguration` object using builder will now generate a `Supplier` which will resolve to a project ID later when `get()` is called on it. This is done by `InternalMetricExporter` when it is lazy loaded.
5. In case the call made to `ServiceOptions.getDefaultProjectId()` (for retrieving default project ID) fails, the `GoogleCloudMetricExporter` implementation will be changed to a noop implementation and the user will be notified of this via a log line. 
 

#### Updated Unit Testing
Added new unit tests to verify the updated logic.

#### Documentation Update
Updated the documentation to explain the deferred approach.


build.gradle
```diff
@@ -120,7 +120,7 @@ subprojects {
 		springOtelVersion = '1.0.0-M8'
 		cloudEventsCoreVersion = '2.4.0'
 		cloudFunctionsFrameworkApiVersion = '1.0.4'
-		opencensusShimVersion = '1.23.1'
+		opencensusShimVersion = '1.4.0'
```
punya, line 123: Why are we downgrading it?

examples/metrics/README.md
```diff
@@ -0,0 +1,42 @@
+### Running the example to export metrics to Google Cloud
```
punya, line 1: This doesn't seem related to the fix -- please move everything related to the new example to a separate PR. Also, if we need two different entry points then we should have two different projects.

exporters/metrics/build.gradle
```diff
@@ -18,15 +18,17 @@ description = 'Cloud Monitoring Exporter for OpenTelemetry'
 dependencies {
```
punya, line 18: It's okay for this PR, but in future please avoid mixing formatting changes with actual code changes.

exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/InternalMetricExporter.java
```diff
@@ -0,0 +1,238 @@
...
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import javax.annotation.Nonnull;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This class encapsulates internal implementation details for exporting metrics for applications
+ * running in Google Cloud environment. This class should not be exposed to the end user for direct
```
punya, line 53: Please remove "applications running in Google Cloud environment" - the metric exporter is expected to cover usage from outside GCE, such as on-premises or in under other cloud providers.

punya (approved):

PR #213, opened 2023-03-23
Improve build time for e2e-test-server

psx95:
This is to improve the build-times as 'shadowJar' generates shadowJars for all sub-projects which is expensive and not required here.

### DIFF
--- a/cloudbuild-e2e-cloud-functions-gen2.yaml
+++ b/cloudbuild-e2e-cloud-functions-gen2.yaml
@@ -18,7 +18,7 @@ steps:
     id: generate-jar
     entrypoint: "gradle"
     timeout: 4m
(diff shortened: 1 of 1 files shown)


cloudbuild-e2e-cloud-functions-gen2.yaml
```diff
@@ -18,7 +18,7 @@ steps:
     id: generate-jar
     entrypoint: "gradle"
     timeout: 4m
-    args: ["shadowJar"]
+    args: [":e2e-test-server:build"]
```
psx95, line 21: build depends on shadowJar, so executing build invokes task 'shadowJar'.

punya (approved):

PR #217, opened 2023-03-29
Make GCPResourceTest environment agnostic

### LINKED ISSUE #216: Isolate environment dependent unit tests for resource detectors
Unit tests for resource detectors depend on the machines they are run on which cause build failures on certain machines. 

For instance, currently `testGCPComputeResourceNotGCP` fails when executing on a GCE VM, but passes locally. This causes build failures on GCE.  

Isolate these tests so that they do not depend on the environment.

psx95:
fixes #216 

The unit test to verify that empty attributes are returned in non GCP environment, assumed that the test environment would be non GCE. This led to failures when running on GCE VMs.

### DIFF
--- a/detectors/resources/build.gradle
+++ b/detectors/resources/build.gradle
@@ -23,5 +23,6 @@ dependencies {
 	testImplementation(testLibraries.assertj)
 	testImplementation(testLibraries.junit)
 	testImplementation(testLibraries.wiremock)
+	testImplementation(testLibraries.mockito)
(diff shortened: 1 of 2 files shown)


punya (commented):
This is meant to be an e2e test, so it's more appropriate to skip it on GCP rather than to insert a mock. Take a look at https://junit.org/junit4/javadoc/4.12/org/junit/Assume.html for one way to do it.

punya (approved):

PR #221, opened 2023-04-10
Add missing otlp exporter dependency

psx95:
Autoconfigure SDK extension is used but has not been configured. In this case, the default otlp exporter must be present on the classpath.

### Testing
 - Ran the resource detection example successfully.


punya:
Why are we adding the OTLP exporter to the classpath rather than changing the default exporter to the Google Cloud ones?

punya (approved):

PR #224, opened 2023-04-11
Add required env variables for publishing metrics

psx95:
The added environmnet variables are required for resource detection in GKE environment. The absence of these variables cause failure to publish metrics.

### DIFF
--- a/examples/autoinstrument/README.md
+++ b/examples/autoinstrument/README.md
@@ -8,8 +8,8 @@ export GOOGLE_CLOUD_PROJECT={your-project}
 
 ./gradlew :examples-autoinstrument:jib --image="gcr.io/$GOOGLE_CLOUD_PROJECT/hello-autoinstrument-java"
 
-kubectl create deployment hello-autoinstrument-java \
-  --image=gcr.io/$GOOGLE_CLOUD_PROJECT/hello-autoinstrument-java
+sed s/%GOOGLE_CLOUD_PROJECT%/$GOOGLE_CLOUD_PROJECT/g \
+ examples/autoinstrument/deployment.yaml | kubectl apply -f -
 
 kubectl expose deployment  hello-autoinstrument-java --type LoadBalancer --port 80 --target-port 8080
 ```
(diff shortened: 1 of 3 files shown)


aabmass (commented):
IMO these should be supplied at the GKE manifest (or kubectl invocation) and not hardcoded in the container

examples/autoinstrument/build.gradle
```diff
@@ -51,6 +51,8 @@ jib {
 	from.image = 'gcr.io/distroless/java-debian10:11'
 	containerizingMode = 'packaged'
 	container.ports = ["8080"]
+	// required for resource detection in GKE environment
```
punya, line 54: This is a sensible default value for `CONTAINER_NAME`.

However, `NAMESPACE` should not be set in the *image*, it should be passed in as part of the *container* definition when needed. This is for two reasons:

1. It's specific to K8s, but a container image can be used in non-K8s environments (such as Cloud Run or a VM with a container runtime).
2. If someone uses this image in K8s but forgets to set `NAMESPACE`, they'll get a misleading result rather than an error.

If we really need to set a value for `NAMESPACE`, we should use something like `unspecified` to signal that someone needs to modify it.

punya (changes requested):

examples/autoinstrument/README.md
```diff
@@ -8,8 +8,8 @@ export GOOGLE_CLOUD_PROJECT={your-project}
 
 ./gradlew :examples-autoinstrument:jib --image="gcr.io/$GOOGLE_CLOUD_PROJECT/hello-autoinstrument-java"
 
-kubectl create deployment hello-autoinstrument-java \
-  --image=gcr.io/$GOOGLE_CLOUD_PROJECT/hello-autoinstrument-java
+sed s/%GOOGLE_CLOUD_PROJECT%/$GOOGLE_CLOUD_PROJECT/g \
```
punya, line 11: Optional: consider using `envsubst` rather than `sed`.

punya (approved):

examples/autoinstrument/README.md
psx95: `envsubst` does seem simpler than using `sed`. A potential downside is that it is not pre-installed on some systems like MacOS - whereas `sed` is.

PR #229, opened 2023-04-13
Add script to run example in docker container

psx95:
This just adds a convenience script that allows users to run auto-instrument example in a local docker container instead of running on GKE.


punya (approved):
Nice!

Issue #215, opened 2023-03-27
Create new release with the shaded auto-exporter


psx95:
After merge of #212, a new release needs to be created that generates both shaded and non-shaded variants of exporter-auto. 

Completing this would also close #161 .

psx95:
Blocked on #237

PR #235, opened 2023-05-01
Configures auto-exporter to publish shaded and non-shaded variants

psx95:
Supports #215 

Currently, if `gradle.properties` have `shadowed` property set to true, then only shadowed variant is published, this change also configures the maven-publish plugin to publish normal Jar along with the shadow-jar. 

The `artifactId` is currently set to be the `archivesBaseName` which based on current configuration for auto-exporter should contain the suffix - `shaded`.

#### Testing
 - Ran `./gradlew clean build`
 - Ran `./gradlew publishToMavenLocal` & verified that the `.m2` repository contained both - shaded and non-shaded JARs.
 
 ### Generated artifacts through local publish 
Below is the tree structure of the generated artifacts (from local `.m2` repo)
 ```
 exporter-auto/
├── 0.1.0-alpha-SNAPSHOT
│   ├── exporter-auto-0.1.0-alpha-SNAPSHOT.jar
│   ├── exporter-auto-0.1.0-alpha-SNAPSHOT-javadoc.jar
│   ├── exporter-auto-0.1.0-alpha-SNAPSHOT.module
│   ├── exporter-auto-0.1.0-alpha-SNAPSHOT.pom
│   ├── exporter-auto-0.1.0-alpha-SNAPSHOT-shaded.jar
│   ├── exporter-auto-0.1.0-alpha-SNAPSHOT-sources.jar
│   └── maven-metadata-local.xml
└── maven-metadata-local.xml
 ```
 
 #### JAR contents for sources JAR - 
 `jar tf exporter-auto-0.1.0-alpha-SNAPSHOT-sources.jar`

```
META-INF/
META-INF/MANIFEST.MF
com/
com/google/
com/google/cloud/
com/google/cloud/opentelemetry/
com/google/cloud/opentelemetry/auto/
com/google/cloud/opentelemetry/auto/GoogleCloudMetricExporterFactory.java
com/google/cloud/opentelemetry/auto/Constants.java
com/google/cloud/opentelemetry/auto/GoogleCloudSpanExporterFactory.java
```

Verified that the sources and javadoc JARs are for the un-shaded variant.

### DIFF
--- a/exporters/auto/build.gradle
+++ b/exporters/auto/build.gradle
@@ -66,19 +66,3 @@ tasks.named('shadowJar') {
 	enableRelocation true
 	relocationPrefix 'shadow'
 }
-
-publishing {
-	publications {
-		maven(MavenPublication) { publication ->
-			from components.java
-		}
-		shadow(MavenPublication) { publication ->
-			from project.shadow.component(publication)
-			artifact sourcesJar
-			artifact javadocJar
-			components.java.withVariantsFromConfiguration(configurations.runtimeElements) {
-				skip()
-			}
-		}
-	}
-}
--- a/exporters/auto/build.gradle
+++ b/exporters/auto/build.gradle
@@ -70,6 +70,9 @@ tasks.named('shadowJar') {
 publishing {
 	publications {
 		maven(MavenPublication) { publication ->
+			from components.java
+		}
+		shadow(MavenPublication) { publication ->
 			from project.shadow.component(publication)
 			artifact sourcesJar
 			artifact javadocJar
--- a/exporters/auto/gradle.properties
+++ b/exporters/auto/gradle.properties
@@ -1,3 +1,2 @@
 release.qualifier=alpha
 release.enabled=true
-shadowed=true


jsuereth (commented):
Can you verify if `-srcs.jar` and `-javadoc.jar` are generated appropriately for the raw component?

psx95:
> Can you verify if `-srcs.jar` and `-javadoc.jar` are generated appropriately for the raw component?

In the same location in the local `.m2` repository, I can see `exporter-auto-0.1.0-alpha-SNAPSHOT-javadoc.jar` & `exporter-auto-0.1.0-alpha-SNAPSHOT-sources.jar`, looked at their contents and they looked appropriate.

psx95:
One other thing that I suspect is a change required in how we determine if an artifact is release version [here](https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/blob/0c1d8401f7735d92d6f40867d6fd4da11c5bf8f5/build.gradle#L72) - Since the artifact ends with `-shaded` here this would result in false, but I am not sure about this. 

Thoughts ?

punya:
> One other thing that I suspect is a change required in how we determine if an artifact is release version [here](https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/blob/0c1d8401f7735d92d6f40867d6fd4da11c5bf8f5/build.gradle#L72) - Since the artifact ends with `-shaded` here this would result in false, but I am not sure about this.
> 
> Thoughts ?

The *artifact* ends with `-shaded` but not the *version* (which is what that check cares about). I would guess that the logic doesn't need to be changed.

punya (approved):

PR #237, opened 2023-05-04
Adds dependency declarations to order tasks

psx95:
The missing dependency declarations result in errors in the release tasks - pointing to possibility of incorrect result depending on order in which these tasks execute.
The release tasks are skipped during normal :build jobs so build passes, but the artifacts snapshot generation for release fails.

Supports #215


psx95:
Not sure why this was not an issue in the previous release, but the shading process for shared-resourcemapping module was altered a little after the last release was made and gradle plugin version was updated too.

punya (approved):

PR #241, opened 2023-05-08
Revert "Adds explicit instruction for shaded release"

psx95:
Reverts GoogleCloudPlatform/opentelemetry-operations-java#240

We do not want shaded version in a different artifact - use classifiers instead.


psx95:
There are likely to be some additional changes that are required to enable classifiers for this artifact.

punya (approved):

PR #245, opened 2023-05-11
Updates exporter-auto readme for shaded-variants

psx95:


jsuereth (approved):

exporters/auto/README.md
```diff
@@ -6,16 +6,31 @@
...
+  <classifier>shaded</classifier>
 </dependency>
 ```
 
+#### Gradle
+```groovy
+implementation "com.google.cloud.opentelemetry:exporter-auto:0.25.1-alpha"
+```
+
+To use the shaded variant with Gradle,
```
punya, line 29: Consider using https://github.com/ryandens/javaagent-gradle-plugin#application-plugin-integration to show how to use the shaded variant as an agent (i.e. what we recommend) rather than adding it to the application classpath.

You can write up similar instructions for Maven by adapting https://stackoverflow.com/questions/14777909/specify-javaagent-argument-with-maven-exec-plugin, but it's a lot more verbose. It's up to you whether you want to do that.

psx95: I will add those instructions.

psx95: Created a separate issue to track this - https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/issues/246

PR #260, opened 2023-08-14
Use faas.instance instead of faas.id in resource detection and mapping

dashpole:
Context: https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/679

This also bumps the e2e test image to pick up the validation for faas.instance.


dashpole:
Just the basic trace scenario is failing, which shouldn't be impacted by this PR.  Resource detection passes

aabmass (approved):

Issue #268, opened 2023-11-13
ops-java-e2e-cloud-functions-gen2 check is continuously failing for all PRs


psx95:
Last PR with passing check for `ops-java-e2e-cloud-functions-gen2` was #256  (merged on Jun 16, 2023). 

Since then, there have been 2 PRs which were merged with failing checks (#257 and #260)

#### Error Description:
The tests failing in this check is - TestBasicTrace. 
Error: 
```
Retrying GetTrace: googleapi: Error 404: Requested entity was not found., notFound
```

dashpole:
As the likely cause of failures, i'll take an initial look at this

dashpole:
My finding is that the span is correctly generated, but doesn't appear for a significant amount of time after the current test finishes.  Increasing the timeout from 10s to 90s [here](https://github.com/GoogleCloudPlatform/opentelemetry-operations-e2e-testing/blob/main/trace_test.go#L75) results in the test passing: https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/pull/270/checks?check_run_id=18730374111.

Given that the test passes consistently for other languages, I suspect it is an issue with the java function, rather than the test framework.  I'm going to try switching from the batch span processor to the synchronous one

PR #270, opened 2023-11-15
Fix cloudfunctions test

dashpole:
Fixes https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/issues/268

It uses a synchronous export of spans to ensure spans are delivered before cloud functions reclaims the CPU.


psx95:
The reasoning behind this change makes sense to me, but I'm still not a 100% sure why the resource detector test passes. It has the same timeout and consistently passed using the same BatchSpanExporter.

Any thoughts on it ?

dashpole:
My theory was that maybe it had something to do with test ordering?  If the basic test always runs last, that could be why it is the one that has the CPU taken away from it.

dashpole:
we can see...  It seems to consistently pass with this change

psx95:
> we can see... It seems to consistently pass with this change

Ok, I think I'm ok to merge this. If the flakiness continues we can dig deeper.

psx95 (approved):

Issue #266, opened 2023-10-13
Contribution of the resource providers to opentelemetry-java-contrib repo


SylvainJuge:
Hi !

I just opened https://github.com/open-telemetry/opentelemetry-java-contrib/issues/1074 in the java contrib repo after discussing the topic in the Java SIG meeting.

Ideally we'd like to have the most common cloud resource provider implementations available, would you be interested in contributing your implementation currently in [detectors/resources](https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/tree/main/detectors/resources) to the [open-telemetry/opentelemetry-java-contrib](https://github.com/open-telemetry/opentelemetry-java-contrib) repo ?

If yes, would you also be interested in being maintainers of this component within otel ? That does not sound like a very strong commitment and you'll benefit from frequent opentelemetry release schedule and automatic version upgrades. I can definitely be a second co-maintainer of this component within otel if needed.

I saw that this project currently relies on the now obsolete `io.opentelemetry:opentelemetry-sdk-extension-resources:1.19.0` dependency, which has now been moved to `opentelemetry-java-contrib` repository, as a consequence replacing this will already create a dependency to the contrib repo.
What I mean here is that ideally a single implementation in the contrib repo could easily be reused here without adding an extra repository as a dependency (just an extra artifact).

SylvainJuge:
For now I see two potential aspects that could be problematic for re-use in OTel-based agents or SDKs:
- resource attributes fetched on demand and one by one, which means more than one HTTP request to the metadata endpoint which could make startup slower.
- currently targets Java 11, whereas upstream opentelemetry is java 8 compatible.

dashpole:
Hey @SylvainJuge, thanks for reaching out. Overall, we would be happy to contribute the resource detector here to the java contrib repo.

One of our original reasons for having it in this repository was to be able to run presubmit integration tests on GCP infrastructure to ensure the resource detectors work as expected. For other languages, like golang, we ended up splitting the detector into a core detection library (https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/tree/main/detectors/gcp), and a contrib entrypoint (https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/detectors/gcp/detector.go).  We can run integration tests on the core library to ensure the detection methods work properly, but the actual detector would live in the upstream contrib repository.  Would something like that work for you?

[...]

PR #271, opened 2023-11-19
Rename GCPResource -> GCPResourceDetctor

psx95:
This is a breaking change for code depending on `detector-resources`.


dashpole:
This seems reasonable.  What is the motivation for the change?

dashpole:
Given https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/issues/266 might move this anyways, lets stick with the current naming for now to avoid breaking users unnecessarily.

psx95:
> Given #266 might move this anyways, lets stick with the current naming for now to avoid breaking users unnecessarily.

Sounds good, but do we have a timeline on #266 ?

dashpole:
we can talk about it at triage today

psx95:
This could be done as part of #266, we can avoid this breaking change in the next release.

PR #272, opened 2023-11-20
Update mappings for generic_task and generic_node

### LINKED ISSUE #265: Map faas.* attributes to generic_task in resource mapping
Implement https://github.com/GoogleCloudPlatform/opentelemetry-operations-python/pull/273 in the resource mapping in this repository.

psx95:
Fixes #265 

This PR: 
 - adds missing faas.* attributes mappings for `generic_task`.
 - adds missing resource mapping for `generic_node`.

### DIFF
--- a/shared/resourcemapping/src/main/java/com/google/cloud/opentelemetry/resource/ResourceTranslator.java
+++ b/shared/resourcemapping/src/main/java/com/google/cloud/opentelemetry/resource/ResourceTranslator.java
@@ -20,6 +20,7 @@ import io.opentelemetry.sdk.resources.Resource;
 import io.opentelemetry.semconv.ResourceAttributes;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 
 /** Translates from OpenTelemetry Resource into Google Cloud's notion of resource. */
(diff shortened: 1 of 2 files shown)


shared/resourcemapping/src/main/java/com/google/cloud/opentelemetry/resource/ResourceTranslator.java
```diff
@@ -126,7 +142,19 @@ public static GcpResource mapResource(Resource resource) {
...
+    }
+  }
+
+  private static GcpResource genericTaskOrNode(Resource resource) {
+    Map<AttributeKey<?>, Object> attrMap = resource.getAttributes().asMap();
+    if ((attrMap.containsKey(ResourceAttributes.SERVICE_NAME)
+            || attrMap.containsKey(ResourceAttributes.FAAS_NAME))
+        && (attrMap.containsKey(ResourceAttributes.SERVICE_INSTANCE_ID)
+            || attrMap.containsKey(ResourceAttributes.FAAS_INSTANCE))) {
+      return mapBase(resource, "generic_task", GENERIC_TASK_LABELS);
```
dashpole, line 155: for generic_task, we also need to ignore service names that start with unknown_service.  See changes here: https://github.com/GoogleCloudPlatform/opentelemetry-operations-python/pull/273/files#diff-9b839eca0dff6b3acf5d664f72cb6ba284ad5fdca5f69ff59f60c15dcb355b04R198.  LMK if you have questions

psx95: I have a couple questions -

1. Is the intention that for `generic_task` you could have both `service.name` and `faas.name` so we ignore `unknown-service` value for `service.name` in favor of a possible value in `faas.name`. However if `faas.name` is also absent, we use the `unknown-service` name value mentioned in `service.name`.

2. IIUC, the code linked will ignore any mr value that starts with `unknown-service` - this is regardless of the label. For instance,  if I have a `namespace` with value `unknown-service-foo`, it will be ignored. Is this intentional ?

dashpole: 1. Yes (although it is unknown_service)
2. Not intentional.  You are welcome to address that in your implementation.

dashpole (approved):

PR #274, opened 2023-11-22
chore: run codeql on PRs

dashpole:
Based on https://github.com/open-telemetry/opentelemetry-go/blob/main/.github/workflows/codeql-analysis.yml and https://github.com/open-telemetry/opentelemetry-java/blob/main/.github/workflows/codeql-daily.yml


psx95 (approved):

psx95 (commented):
~I think it needs to be changed a bit since we do not use kotlin~

Misread the error - probably because this PR does not modify any source code.

PR #276, opened 2023-12-01
Add new module for resource-detector support library

psx95:
Separates out resource detection module into a support-library (detector-resources-support) and a mapper (detector-resources). 

The mapper uses the support-library to map OTel specific resources to GCP specific resources. The support-library contains logic to detect the GCP specific resources and provided appropriate attributes for the detected resource.

The logic split has been done keeping the following in mind - 

1. The support library would not contain any Google-specific dependency (e.g. a dependency on any GCP SDK). 
2. The support library would not contain any OpenTelemetry specific dependency. 
3. The mapper library could potentially go upstream in #266 and would take a dependency on the support-library.

Supports #266


detectors/resources-support/src/main/java/com/google/cloud/opentelemetry/detectors/EnvironmentVariables.java
psx95, line 1: TODO: Remove public access modifier from this class.

detectors/resources-support/src/main/java/com/google/cloud/opentelemetry/detectors/CloudLocationUtil.java
```diff
@@ -0,0 +1,86 @@
...
+ */
+package com.google.cloud.opentelemetry.detectors;
+
+import java.util.Optional;
+
+/**
+ * Contains utility functions to retrieve appropriate zone and regions representing the location of
+ * certain cloud resources.
+ */
+final class CloudLocationUtil {
```
jsuereth, line 24: I'd just go fix the GCPMetaDataConfig class to have these methods on it, rather than in a util.

detectors/resources-support/src/main/java/com/google/cloud/opentelemetry/detectors/GCPPlatformDetector.java
```diff
@@ -0,0 +1,115 @@
...
+    this.metadataConfig = GCPMetadataConfig.DEFAULT_INSTANCE;
+    this.environmentVariables = EnvironmentVariables.DEFAULT_INSTANCE;
+  }
+
+  // Detects the GCP platform on which the application is running
+  public DetectedPlatform detectPlatform() {
+    if (!isRunningOnGcp()) {
+      return generateDetectedPlatform(SupportedPlatform.UNKNOWN_PLATFORM);
+    }
+    Function<EnvironmentVariables, Optional<SupportedPlatform>> detectGKE =
```
jsuereth, line 44: I'm not sure how this is more readable than an if/else statement.  Is there a reason to construct all these thunks?

psx95: In this case, the detection functions ended up being one-liners, so if-else seems like a viable replacement, but in case the platform detection was a more involved process, I feel this approach would be more readable one. 

IMO, it also streamlines the process of running detection one-by-one on each platform.

detectors/resources-support/src/main/java/com/google/cloud/opentelemetry/detectors/CloudLocationUtil.java
psx95: This makes sense. I will remove this class.

detectors/resources-support/src/main/java/com/google/cloud/opentelemetry/detectors/GCPPlatformDetector.java
```diff
@@ -0,0 +1,115 @@
...
+                : Optional.empty();
+    Function<EnvironmentVariables, Optional<SupportedPlatform>> detectGAE =
+        environmentVariables ->
+            environmentVariables.get("GAE_SERVICE") != null
+                ? Optional.of(SupportedPlatform.GOOGLE_APP_ENGINE)
+                : Optional.empty();
+
+    // Order of detection functions matters here
+    Stream<Function<EnvironmentVariables, Optional<SupportedPlatform>>> platforms =
+        Stream.of(detectGKE, detectGCR, detectGCF, detectGAE);
```
jsuereth, line 68: Shouldn't GCF be before GCR?

detectors/resources-support/src/main/java/com/google/cloud/opentelemetry/detectors/GoogleAppEngine.java
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  // for testing only
+  GoogleAppEngine(EnvironmentVariables environmentVariables, GCPMetadataConfig metadataConfig) {
+    this.environmentVariables = environmentVariables;
+    this.metadataConfig = metadataConfig;
+    this.availableAttributes = prepareAttributes();
+  }
+
+  private Map<String, Optional<String>> prepareAttributes() {
```
jsuereth, line 47: Returning `Map<String, Optional<String>>` is fairly uncommon in Java.  In functional (newer) languages, like Kotlin or Scala, you'd expect `Map` to support a `get` that returns an `Optional` value, denoting non-existence.

Here, I'd suggest still using `Map<String,String>` and on the read side doing:


```java
Optional.ofNullable(map.get(key))
```

This will still let you use your `ifPresent` trick and is less prone to runtime exceptions than what you're doing now (in the event your keys do not exist in the map).

detectors/resources-support/src/main/java/com/google/cloud/opentelemetry/detectors/GCPPlatformDetector.java
psx95: I agree with the additional pressure on GC point. Personally, I do not prefer if-else-if ladders but as you mentioned detection logic is unlikely to be changed - so I will update the code to remove the additional Functions.

detectors/resources-support/src/main/java/com/google/cloud/opentelemetry/detectors/GCPPlatformDetector.java
psx95: I did not look at the spec again, but I was going by the code we already had - in which [we checked for GCR before GCF](https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/blob/de1d4b4dbb25503afe271d28f68f20c81740d0b8/detectors/resources/src/main/java/com/google/cloud/opentelemetry/detectors/GCPResource.java#L65).

Is there a reason why GCF should be checked before GCR ?

detectors/resources-support/src/main/java/com/google/cloud/opentelemetry/detectors/GoogleAppEngine.java
psx95: Sounds good 👍🏻 

I agree - the main reason behind using `Optional` was the `ifPresent` trick.

detectors/resources-support/src/main/java/com/google/cloud/opentelemetry/detectors/DetectedPlatform.java
```diff
@@ -0,0 +1,37 @@
...
+   */
+  GCPPlatformDetector.SupportedPlatform getSupportedPlatform();
+
+  /**
+   * Method to retrieve the attributes associated with the compute platform on which the application
+   * is running as key-value pairs.
+   *
+   * @return a {@link Map} of attributes specific to the underlying compute platform.
+   */
+  Map<String, String> getAttributes();
```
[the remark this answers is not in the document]
psx95: It helped in writing simpler mocks and made the code testable. I don't think the overhead of additional constants is too much especially since some of them are getting re-used and Java string pool usually does a good job in re-using string literals. 

@jsuereth can maybe comment here on the trade-off.

detectors/resources-support/src/main/java/com/google/cloud/opentelemetry/detectors/AttributeKeys.java
```diff
@@ -0,0 +1,64 @@
...
+  // GCE Attributes
+  public static final String GCE_PROJECT_ID = AttributeKeys.PROJECT_ID;
+  public static final String GCE_AVAILABILITY_ZONE = AttributeKeys.AVAILABILITY_ZONE;
+  public static final String GCE_CLOUD_REGION = AttributeKeys.CLOUD_REGION;
+  public static final String GCE_INSTANCE_ID = AttributeKeys.INSTANCE_ID;
+  public static final String GCE_INSTANCE_NAME = AttributeKeys.INSTANCE_NAME;
+  public static final String GCE_MACHINE_TYPE = AttributeKeys.MACHINE_TYPE;
+
+  // GKE Attributes
+  public static final String GKE_POD_NAME = "gke_pod_name";
```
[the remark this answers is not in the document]
psx95: The existing detector did do this. 

I saw resource attributes present in OTel semconv like `K8S_POD_NAME`, `K8S_NAMESPACE_NAME` etc which are mapped to these - why do you say we shouldn't detect these?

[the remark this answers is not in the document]
dashpole: If multiple detectors are used that detect a resource attribute, they can result in a collision and an error to users.  It would be preferable to have a k8s resource detector that does this for all kubernetes platforms.

detectors/resources-support/src/main/java/com/google/cloud/opentelemetry/detectors/DetectedPlatform.java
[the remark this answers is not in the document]
dashpole: I'm not worried about the overhead, as resource attributes are a one-time cost.  Does this make it more complicated for the user of this library?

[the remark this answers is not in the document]
psx95: The only user of this library is gonna be the upstream detector (detector-resources module) - The trade-off we make here is choosing between calling resource specific getter functions and making a get call on the Attributes map. 

While a getter function would be marginally simpler on the consumer side, it would mean exposing all the compute specific classes on the consumer side - since there would be no way to unify the individual GCP compute classes via an interface (since Getter methods would be different for each platform).

IMO, the code on the consumer side is not too complicated to read and maintain.

detectors/resources-support/src/main/java/com/google/cloud/opentelemetry/detectors/AttributeKeys.java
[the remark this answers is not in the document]
psx95: Discussed offline - will remove K8s specific attributes and match the detected attributes across platforms to be same as the go detector.

detectors/resources-support/src/main/java/com/google/cloud/opentelemetry/detectors/DetectedPlatform.java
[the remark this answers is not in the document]
dashpole: ok, that sounds reasonable.

dashpole (approved):

SylvainJuge (approved):

PR #285, opened 2024-01-31
Use logstash-logback-encoder version that works with spring boot 2

aabmass:
This is a port over of https://github.com/GoogleCloudPlatform/java-docs-samples/pull/8984/ to this repo.

The breakage was actually caused by upgrading logstash-encoder to 7.4 because [it does not support logback versions prior to 1.3.0](https://github.com/logfellow/logstash-logback-encoder#:~:text=Support%20for%20logback%20versions%20prior%20to%201.3.0%20was%20removed%20in%20logstash%2Dlogback%2Dencoder%207.4.). Spring Boot 2 uses an [older logback version](https://docs.spring.io/spring-boot/docs/2.7.18/reference/html/dependency-versions.html).

### DIFF
--- a/examples/instrumentation-quickstart/pom.xml
+++ b/examples/instrumentation-quickstart/pom.xml
@@ -82,7 +82,9 @@
         <dependency>
             <groupId>net.logstash.logback</groupId>
             <artifactId>logstash-logback-encoder</artifactId>
-            <version>7.4</version>
+              <!-- Cannot be updated until logback is updated to 1.3+, probably in the next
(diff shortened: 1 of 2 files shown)


aabmass:
I will upgrade to spring 3, gradle, add CI, etc. later on. Just want to get things in a working state

psx95 (approved):

psx95:
I approved but let's switch this example to use gradle as well, not being on gradle is the reason why #243 has not been approved.

PR #291, opened 2024-02-08
Upgrade examples/instrumentation-quickstart to use Spring Boot 3

aabmass:


examples/instrumentation-quickstart/build.gradle
```diff
@@ -25,6 +25,10 @@ group = 'com.example'
 version = '0.0.1-SNAPSHOT'
 description = 'OpenTelemetry instrumentation quickstart using Spring Boot'
 
+java {
+	sourceCompatibility = '17'
+}
+
```
psx95, line 31: Since the minimum compatibility is different from other examples, could you mention the minimum compatibility in the README for this example ?

psx95 (changes requested):
Nit: A small readme change regarding minimum compatibility would be better.

aabmass:
Decided against doing this, and will just leave it on Spring Boot 2 for now. There are some benefits to using Spring Boot 3 and Java 17, but I've already done the hard work of working around Java 8 and old logback verisons

PR #296, opened 2024-02-09
Update TraceTranslator.java

### LINKED ISSUE #295: Trace Exporter sets invalid parent span IDs for root spans
Root spans should have the parent span ID left empty but Trace Exporter is setting them to `0000000000000000` which is not a valid span ID and shows up in Cloud Trace as `(Missing span ID 0000000000000000)`.

image

I believe the issue is in [exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceTranslator.java](https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/blob/70f129bfaf85e2bdd3975915f06ce9aa4e2b6806/exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceTranslator.java#L123)

```java
if (spanData.getParentSpanId() != null) {
  spanBuilder.setParentSpanId(spanData.getParentSpanId());
}
```

Instead of checking to see if the parent span ID  is not null the check should be for if the parent context is valid:

```java
if (spanData.getParentSpanContext().isValid()) {
  spanBuilder.setParentSpanId(spanData.getParentSpanId());
}
```

After making that change to a local copy of the TraceTranslator.java I no longer see the `(Missing span ID 0000000000000000)` spans and the root span has the parent span ID empty.

image

WadeGulbrandsen:
Check to see if parent context is valid instead of parent span ID not null

Fixes #295

### DIFF
--- a/exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceTranslator.java
+++ b/exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceTranslator.java
@@ -120,7 +120,7 @@ class TraceTranslator {
       spanBuilder.setEndTime(toTimestampProto(end));
     }
     spanBuilder.setLinks(toLinksProto(spanData.getLinks(), spanData.getTotalRecordedLinks()));
-    if (spanData.getParentSpanId() != null) {
+    if (spanData.getParentSpanContext().isValid()) {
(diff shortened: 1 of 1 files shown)


psx95 (approved):
Thank you for the fix.

PR #297, opened 2024-02-15
Add OTLP trace with ADC example

damemi:
Cobbled this together following docs from:

* https://opentelemetry.io/docs/languages/java/exporters/
* https://github.com/googleapis/google-auth-library-java
* https://github.com/googleapis/google-auth-library-java/blob/main/samples/snippets/src/main/java/AuthenticateExplicit.java
* https://cloud.google.com/java/docs/reference/google-auth-library/latest/com.google.auth.oauth2.GoogleCredentials
* https://javadoc.io/static/io.opentelemetry/opentelemetry-exporter-otlp-http-trace/1.14.0/io/opentelemetry/exporter/otlp/http/trace/OtlpHttpSpanExporterBuilder.html

Same idea as https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/800, https://github.com/GoogleCloudPlatform/opentelemetry-operations-python/pull/307, and https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/pull/676

### DIFF
--- a/examples/otlptrace/src/main/java/com/google/cloud/opentelemetry/example/otlptrace/OTLPTraceExample.java
+++ b/examples/otlptrace/src/main/java/com/google/cloud/opentelemetry/example/otlptrace/OTLPTraceExample.java
@@ -15,14 +15,11 @@
  */
 package com.google.cloud.opentelemetry.example.otlptrace;
 
-import com.google.auth.oauth2.GoogleCredentials;
 import io.opentelemetry.api.trace.Span;
 import io.opentelemetry.context.Scope;
-import io.opentelemetry.exporter.otlp.http.trace.OtlpHttpSpanExporter;
 import io.opentelemetry.sdk.OpenTelemetrySdk;
+import io.opentelemetry.sdk.autoconfigure.AutoConfiguredOpenTelemetrySdk;
 import io.opentelemetry.sdk.common.CompletableResultCode;
-import io.opentelemetry.sdk.trace.SdkTracerProvider;
-import io.opentelemetry.sdk.trace.export.BatchSpanProcessor;
 import java.io.IOException;
 import java.util.Random;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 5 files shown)


psx95:
Is this different from [this example](https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/blob/main/examples/trace/src/main/java/com/google/cloud/opentelemetry/example/trace/TraceExporterExample.java) which is already present ?

It seems similar.

psx95:
> Is this different from [this example](https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/blob/main/examples/trace/src/main/java/com/google/cloud/opentelemetry/example/trace/TraceExporterExample.java) which is already present ?
> 
> It seems similar.

Ah, nvm it uses OTLP Span exporter instead of Google Cloud Exporter.

psx95 (commented):
The sample looks good. A couple of follow-ups: 
 - Should we move it to the existing `examples` folder ? The proposed example is placed in a folder called `example`. 
  - Could we add a corresponding gradle file along with a small Readme with instructions to build and run this example so that it remains consistent with the rest of the examples.

psx95 (changes requested):
Looks good to me! Thanks @damemi ! 👍🏻 

Following from our offline discussion, two pending action items prior to merging - 
 - Adding `-Dotel.exporter.otlp.protocol=http/protobuf'` to the `autoconf_config` in `build.gradle`
 -  Adding a README file under `examples/otlptrace` to give an overview of this example and instructions on running it.

damemi:
Thanks for all your help @psx95! I added the changes we talked about and tested this and it's working now 🎉

examples/otlptrace/src/main/java/com/google/cloud/opentelemetry/example/otlptrace/OTLPTraceExample.java
```diff
@@ -0,0 +1,116 @@
...
+        credentials.refreshIfExpired();
+        OtlpHttpSpanExporterBuilder builder =
+            ((OtlpHttpSpanExporter) exporter)
+                .toBuilder()
+                    .addHeader(
+                        "Authorization", "Bearer " + credentials.getAccessToken().getTokenValue());
+
+        return builder.build();
+      } catch (IOException e) {
+        System.out.println("error");
```
psx95, line 64: Nit: Could be better to log the underlying error message as well 

```
System.out.println("error:  " + e.getMessage());
```

psx95 (approved):

PR #298, opened 2024-02-27
feat: simplify logging in instrumentation quickstart

dashpole:
Same as https://github.com/GoogleCloudPlatform/golang-samples/pull/3801

Tested manually

### DIFF
--- a/examples/instrumentation-quickstart/docker-compose.yaml
+++ b/examples/instrumentation-quickstart/docker-compose.yaml
@@ -21,7 +21,6 @@ services:
       - OTEL_EXPORTER_OTLP_ENDPOINT=http://otelcol:4317
       - OTEL_SERVICE_NAME=otel-quickstart-spring-boot
       - OTEL_METRIC_EXPORT_INTERVAL=5000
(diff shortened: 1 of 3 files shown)


examples/instrumentation-quickstart/docker-compose.yaml
```diff
@@ -21,7 +21,6 @@ services:
       - OTEL_EXPORTER_OTLP_ENDPOINT=http://otelcol:4317
       - OTEL_SERVICE_NAME=otel-quickstart-spring-boot
       - OTEL_METRIC_EXPORT_INTERVAL=5000
-      - GOOGLE_CLOUD_PROJECT
```
psx95, line 24: Interesting - was this not supposed to have a value ?

psx95 (approved):

examples/instrumentation-quickstart/docker-compose.yaml
dashpole: not sure.  Maybe it implicitly means "pass through the GOOGLE_CLOUD_PROJECT env var"?

PR #302, opened 2024-03-05
Update links to the upstream detector

psx95:


detectors/resources-support/README.md
```diff
@@ -2,7 +2,7 @@
 
 [![Maven Central][maven-image]][maven-url]
 
-This library provides detection support for supported GCP environments and is intended to be consumed by the [GCP OpenTelemetry Detector](../resources/README.md).
+This library provides detection support for supported GCP environments and is intended to be consumed by the [GCP OpenTelemetry Detector](https://github.com/open-telemetry/opentelemetry-java-contrib/tree/main/gcp-resources).
```
dashpole, line 5: prefer linking to a specific commit, in-case it is moved.  Here and below

dashpole (approved):

PR #314, opened 2024-03-25
Add resource attributes as metric labels

### LINKED ISSUE #313: Service name does not appear in metric labels when using auto-instrumentation
The metrics produced by auto-instrumentation do not show service name in the metric labels in Metrics Explorer in GCP Console.

The service name is set using the JVM Args - `'-Dotel.service.name=<service_name>',` and shows up in the trace attributes, but does not show up in the metric labels.

psx95:
#### Feature Addition
 - Adds user configurable option to attach `service.name`, `service.namespace` and `service.id` resource attributes to metric labels. 
 - This can be disabled by passing the provided `NO_RESOURCE_ATTRIBUTES` as the resource filter to the `MetricConfiguration` object.
 - The default metric configuration has been changed to use the `DEFAULT_RESOURCE_ATTRIBUTES_FILTER`. 

Note: This is similar to how the [Google Cloud Monitoring go exporter works](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/2be0414d564ce62759e2da63615b27103ab950d3/exporter/metric/option.go#L114).

#### Testing
 - When running the auto-instrumentation-example, `service_name` appears under metric labels on GCP console with the expected value.
 - When running the metric export example, `service_name` appears with `unknown_resource` since the example does not include resource detection.

#### Deprecation
 - Deprecates the current publicly available `AggregateByLabelMetricTimeSeriesBuilder` constructor in favor of one that also accepts a resource attribute filter in form of a predicate. The existing constructor uses the `NO_RESOURCE_ATTRIBUTES` to keep it in-line with the existing functionality.

fixes #313

### DIFF
--- a/exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/AggregateByLabelMetricTimeSeriesBuilder.java
+++ b/exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/AggregateByLabelMetricTimeSeriesBuilder.java
@@ -27,17 +27,20 @@ import com.google.monitoring.v3.TimeSeries;
 import com.google.monitoring.v3.TypedValue;
 import io.opentelemetry.api.common.AttributeKey;
 import io.opentelemetry.api.common.Attributes;
+import io.opentelemetry.api.common.AttributesBuilder;
 import io.opentelemetry.sdk.common.InstrumentationScopeInfo;
 import io.opentelemetry.sdk.metrics.data.DoublePointData;
 import io.opentelemetry.sdk.metrics.data.HistogramPointData;
 import io.opentelemetry.sdk.metrics.data.LongPointData;
 import io.opentelemetry.sdk.metrics.data.MetricData;
 import io.opentelemetry.sdk.metrics.data.PointData;
+import io.opentelemetry.sdk.resources.Resource;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
 /**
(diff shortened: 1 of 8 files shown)


exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/AggregateByLabelMetricTimeSeriesBuilder.java
```diff
@@ -55,10 +58,20 @@ public final class AggregateByLabelMetricTimeSeriesBuilder implements MetricTime
...
+  private final Predicate<AttributeKey<?>> resourceAttributeFilter;
 
+  @Deprecated
   public AggregateByLabelMetricTimeSeriesBuilder(String projectId, String prefix) {
     this.projectId = projectId;
     this.prefix = prefix;
+    this.resourceAttributeFilter = MetricConfiguration.NO_RESOURCE_ATTRIBUTES;
+  }
+
+  public AggregateByLabelMetricTimeSeriesBuilder(
```
dashpole, line 70: Does this mean users are required to pass a predicate in when they construct an exporter?  Or is this function not really meant to be used directly by users?

psx95: IMO, this function is not meant to be used directly by the customers. 

However, because it was already public, I added a second constructor instead of updating the old one to avoid breaking changes.

dashpole: Sounds good

dashpole (approved):

jsuereth (approved):

PR #317, opened 2024-03-29
Update resource detection sample

psx95:
- Updates Readme to add missing prerequisites.
 - Attempts to clarify the intention of the resource example.

Once this is merged, the [README for the upstream GCP Detector](https://github.com/open-telemetry/opentelemetry-java-contrib/tree/main/gcp-resources) can be updated to clarify that recommended way for resource detection is via `autoconfiguration` and we can link to this example for further clarification.

### DIFF
--- a/examples/resource/README.md
+++ b/examples/resource/README.md
@@ -1,13 +1,62 @@
 # Resource detection Example
 
-An example application that shows what resource attributes will be detected.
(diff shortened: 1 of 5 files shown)


examples/resource/README.md
```diff
@@ -1,13 +1,62 @@
 # Resource detection Example
 
-An example application that shows what resource attributes will be detected.
+An example application that shows what resource attributes will be detected by the [GCP Resource Detector](https://github.com/open-telemetry/opentelemetry-java-contrib/tree/main/gcp-resources) and how the [Autoconfigure Resource Provider SPI](https://github.com/open-telemetry/opentelemetry-java/tree/main/sdk-extensions/autoconfigure#resource-provider-spi) automatically *attaches* the detected resource attributes to the generated telemetry.
```
dashpole, line 3: It would be nice to explicitly state at the top that this does not generate any telemetry, as most users trying out an example probably expect to find traces, metrics, or logs.

dashpole (approved):

PR #318, opened 2024-04-01
Add support for createServiceTimeSeries

### LINKED ISSUE #307: Add support for CreateServiceTimeSeries in metrics exporter
Similar to https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/820.

Add configuration to the metrics exporter to use CreateServiceTimeSeries instead of createtimeseries: https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/blob/398e552233be9261351706dbb0eaa0eca65fb5bd/exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/CloudMetricClientImpl.java#L40

Do not create metric descriptors when using createserivcetimeseries

psx95:
Fixes #307

### DIFF
--- a/exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/MetricConfiguration.java
+++ b/exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/MetricConfiguration.java
@@ -227,6 +227,18 @@ public abstract class MetricConfiguration {
     public abstract Builder setMetricServiceEndpoint(String endpoint);
 
     /**
+     * Sets the {@link MetricConfiguration} to configure the exporter to write metrics via {@link
+     * com.google.cloud.monitoring.v3.MetricServiceClient#createServiceTimeSeries(String, List)}
+     * method. By default, this is false.
+     *
+     * @param useServiceTimeSeries a boolean indicating whether to use {@link
+     *     com.google.cloud.monitoring.v3.MetricServiceClient#createServiceTimeSeries(String, List)}
+     *     method for writing metrics to Google Cloud Monitoring.
+     * @return this
+     */
+    public abstract Builder setUseServiceTimeSeries(boolean useServiceTimeSeries);
+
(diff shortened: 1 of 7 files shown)


exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/MetricConfiguration.java
```diff
@@ -215,6 +226,18 @@ public final Builder setProjectId(String projectId) {
...
+     * Sets the {@link MetricConfiguration} to configure the exporter to write metrics via {@link
+     * com.google.cloud.monitoring.v3.MetricServiceClient#createServiceTimeSeries(String, List)}
+     * method. By default, this is false.
+     *
+     * @param useServiceTimeSeries a boolean indicating whether to use {@link
+     *     com.google.cloud.monitoring.v3.MetricServiceClient#createServiceTimeSeries(String, List)}
+     *     method for writing metrics to Google Cloud Monitoring.
+     * @return this
+     */
+    public abstract Builder setUseServiceTimeSeries(boolean useServiceTimeSeries);
```
dashpole, line 239: Not sure if this is a java pattern, but does this need an argument?  I.e. setUserServiceTimeSeries()?

psx95: I think it is common to pass a  parameter when using `setFoo` methods.

I would think if you wish to avoid passing the parameter,  then the method should be called something like `enableFoo`. Consequently, you would then also need to add `disableFoo` - but in our current situation this would make the builder pattern inconsistent.

dashpole: I think the current approach is acceptable, then.

dashpole (approved):

PR #323, opened 2024-04-10
Add auth support for gRPC

psx95:
It is possible to generate traces using gRPC as well.

### Testing
 - `gradle run`

### DIFF
--- a/examples/otlptrace/src/main/java/com/google/cloud/opentelemetry/example/otlptrace/OTLPTraceExample.java
+++ b/examples/otlptrace/src/main/java/com/google/cloud/opentelemetry/example/otlptrace/OTLPTraceExample.java
@@ -49,6 +49,7 @@ public class OTLPTraceExample {
   }
 
   // Modifies the span exporter initially auto-configured using environment variables
+  // Note: This adds static authorization headers which are set only at initialization time.
   private static SpanExporter addAuthorizationHeaders(
       SpanExporter exporter, GoogleCredentials credentials) {
     if (exporter instanceof OtlpHttpSpanExporter) {
--- a/examples/otlptrace/src/main/java/com/google/cloud/opentelemetry/example/otlptrace/OTLPTraceExample.java
+++ b/examples/otlptrace/src/main/java/com/google/cloud/opentelemetry/example/otlptrace/OTLPTraceExample.java
@@ -50,6 +50,7 @@ public class OTLPTraceExample {
 
   // Modifies the span exporter initially auto-configured using environment variables
   // Note: This adds static authorization headers which are set only at initialization time.
+  // This will stop working after the token expires, since the token is not refreshed.
   private static SpanExporter addAuthorizationHeaders(
       SpanExporter exporter, GoogleCredentials credentials) {
     if (exporter instanceof OtlpHttpSpanExporter) {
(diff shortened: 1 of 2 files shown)


examples/otlptrace/src/main/java/com/google/cloud/opentelemetry/example/otlptrace/OTLPTraceExample.java
```diff
@@ -63,6 +64,18 @@ private static SpanExporter addAuthorizationHeaders(
       } catch (IOException e) {
         System.out.println("error:" + e.getMessage());
       }
+    } else if (exporter instanceof OtlpGrpcSpanExporter) {
+      try {
+        credentials.refreshIfExpired();
+        OtlpGrpcSpanExporterBuilder builder =
```
dashpole, line 70: Do we have the ability to directly pass a credentials option?  https://github.com/grpc/grpc-java/blob/34e241a60e7dd54195daaec3a16decae0dd702cc/api/src/main/java/io/grpc/CallOptions.java#L137

It might be nice to open an issue upstream if that isn't possible today

psx95: I think gRPC's concept of `CallOptions` is not exposed to the end user in OTel Java's OTLP exporters.

That being said, I think closest analog to `WithCallCredentials` is the concept of an [`Authenticator`](https://github.com/open-telemetry/opentelemetry-java/blob/9e12dd60c457ddbe0f5a178fc9790f7a22c1e926/exporters/common/src/main/java/io/opentelemetry/exporter/internal/auth/Authenticator.java#L19) which can be applied to both OTLPGrpcExporter and OTLPHttpExporter. 

However this class is in the internal (experimental) package currently.

dashpole: Cool.  It might be worth opening an issue upstream to see if they can accept CallCredentials in some form.

examples/otlptrace/src/main/java/com/google/cloud/opentelemetry/example/otlptrace/OTLPTraceExample.java
```diff
@@ -63,6 +64,18 @@ private static SpanExporter addAuthorizationHeaders(
       } catch (IOException e) {
         System.out.println("error:" + e.getMessage());
       }
+    } else if (exporter instanceof OtlpGrpcSpanExporter) {
+      try {
+        credentials.refreshIfExpired();
```
dashpole, line 69: this will only refresh the first time, right?  After they expire, the exporter will stop working.

dashpole: If that is the case, please add a big warning comment

psx95: Yes, that is the case, I will add a warning.

examples/otlptrace/src/main/java/com/google/cloud/opentelemetry/example/otlptrace/OTLPTraceExample.java
```diff
@@ -49,6 +49,7 @@ private static OpenTelemetrySdk setupTraceExporter() throws IOException {
   }
 
   // Modifies the span exporter initially auto-configured using environment variables
+  // Note: This adds static authorization headers which are set only at initialization time.
```
dashpole, line 52:
```suggestion
  // Note: This adds static authorization headers which are set only at initialization time.
  // This will stop working after the token expires, since the token is not refreshed.
```

dashpole (approved):

PR #327, opened 2024-04-16
Add otlp with spring boot sample

psx95:
The sample demonstrates exporting OTLP traces from a Spring Boot application along with use of Google Authentication with auto-refreshing credentials.

### Testing
 - Credential refresh tested using the long running client on GKE.


dashpole (approved):

dashpole:
(non-blocking) Did you have any thoughts/ideas for how to make auth easier for users?  Any changes we could propose upstream, or helper libraries we could offer from this repo?  It would be good to open any tracking issues for potential improvements so we don't forget

psx95:
> (non-blocking) Did you have any thoughts/ideas for how to make auth easier for users? Any changes we could propose upstream, or helper libraries we could offer from this repo? It would be good to open any tracking issues for potential improvements so we don't forget

I will think it over and document any potential improvements I can come up with.

PR #328, opened 2024-04-16
Switch to ConcurrentMap for thread-safety

psx95:
Switching to concurrentHashMap since the performance impact is minimal and we gain thread-safety when accessing/updating the attribute cache.


dashpole (approved):

dashpole:
Does this fix any open issues?

psx95:
> Does this fix any open issues?

No, it is just an enhancement for standalone usages of this artifact, which are rare.

PR #329, opened 2024-04-17
Update quickstart sample to output correct severity strings

aabmass:
Similar to https://github.com/GoogleCloudPlatform/golang-samples/pull/3999, this
- makes the collector parse severity strings the same way the Cloud Logging agent does
- outputs the severity levels matching the [LogSeverity](https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#LogSeverity) enum. SLF4J levels are compatible except `TRACE` and `WARN`

manually tested by adding log statements and inspecting
```console
{"timestamp":"2024-04-17T17:54:22.598Z","@version":"1","message":"warn - handle /multi request with subRequests=7","logger_name":"com.example.demo.MultiController","thread_name":"reactor-http-epoll-2","logging.googleapis.com/trace":"661e81d941fc0bc07c533136df6edd30","logging.googleapis.com/spanId":"f99ecaa4b31acc1a","logging.googleapis.com/trace_sampled":true,"severity":"WARNING"}
```

```console
{"timestamp":"2024-04-17T17:54:22.598Z","@version":"1","message":"trace - handle /multi request with subRequests=7","logger_name":"com.example.demo.MultiController","thread_name":"reactor-http-epoll-2","logging.googleapis.com/trace":"661e81d941fc0bc07c533136df6edd30","logging.googleapis.com/spanId":"f99ecaa4b31acc1a","logging.googleapis.com/trace_sampled":true,"severity":"DEBUG"}
```

### DIFF
--- a/examples/instrumentation-quickstart/src/main/resources/logback.xml
+++ b/examples/instrumentation-quickstart/src/main/resources/logback.xml
@@ -25,15 +25,16 @@
           {
             "logging.googleapis.com/trace": "%mdc{trace_id}",
             "logging.googleapis.com/spanId": "%mdc{span_id}",
-            "logging.googleapis.com/trace_sampled": "#asBoolean{%replace(%mdc{trace_flags}){'01', 'true'}}"
+            "logging.googleapis.com/trace_sampled": "#asBoolean{%replace(%mdc{trace_flags}){'01', 'true'}}",
+            "severity": "%replace(%replace(%level){'WARN', 'WARNING'}){'TRACE', 'DEBUG'}"
           }
         </pattern>
       </provider>
 
       <!-- Rename fields for the Logging agent -->
       <fieldNames>
-        <level>severity</level>
         <timestamp>timestamp</timestamp>
+        <level>[ignore]</level>
         <levelValue>[ignore]</levelValue>
       </fieldNames>
 
--- a/examples/instrumentation-quickstart/otel-collector-config.yaml
+++ b/examples/instrumentation-quickstart/otel-collector-config.yaml
@@ -35,6 +35,18 @@ receivers:
           layout: '%Y-%m-%dT%H:%M:%S.%fZ'
         severity:
           parse_from: body.severity
+          preset: none
+          # parse minimal set of severity strings that Cloud Logging explicitly supports
+          # https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#LogSeverity
+          mapping:
+            debug: debug
+            info: info
+            info3: notice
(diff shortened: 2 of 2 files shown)


dashpole (approved):

examples/instrumentation-quickstart/src/main/resources/logback.xml
```diff
@@ -25,15 +25,16 @@
           {
             "logging.googleapis.com/trace": "%mdc{trace_id}",
             "logging.googleapis.com/spanId": "%mdc{span_id}",
-            "logging.googleapis.com/trace_sampled": "#asBoolean{%replace(%mdc{trace_flags}){'01', 'true'}}"
+            "logging.googleapis.com/trace_sampled": "#asBoolean{%replace(%mdc{trace_flags}){'01', 'true'}}",
+            "severity": "%replace(%replace(%level){'WARN', 'WARNING'}){'TRACE', 'DEBUG'}"
```
aabmass, line 29: @dashpole any thoughts on this? Do you find it to be decently readable or hacky?

dashpole: definitely seems hacky to me, but I would say the same about the trace_sampled line.  I'm not familiar with the operators (# or %)

dashpole: not sure if someone used to logback would be more familiar

aabmass: We are using the logstash encoder for JSON support which has a configurable [JSON provider](https://github.com/logfellow/logstash-logback-encoder?tab=readme-ov-file#pattern-json-provider.). The [conversion specifiers](https://logback.qos.ch/manual/layouts.html#conversionWord) are mostly provided by logback.

> not sure if someone used to logback would be more familiar

+1

psx95:
The text "warn" and "trace" in the `message` field was manually added for debugging only, correct ? 

```
,"message":"warn - handle /multi request with subRequests=7",
```

examples/instrumentation-quickstart/src/main/resources/logback.xml
psx95: I think we should create a tracking ticket for this work,  I think it might require some dedicated effort to figure out the solution to this.

What might help in the meantime could be comments within the XML explaining the rationale behind this change because it is not straightforward for the reader.

aabmass:
Closing in favor of https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/pull/330

PR #330, opened 2024-04-18
Change OTel quickstart to use log4j2 instead of logback

aabmass:
Log4j2's `JsonTemplateLayout` supports most of the Cloud Logging special JSON keys directly through the [`GcpLayout` event template](https://logging.apache.org/log4j/2.x/manual/json-template-layout.html#event-templates). I still needed to map the OTel MDC keys for trace correlation, but am planning to contribute [this upstream](https://github.com/apache/logging-log4j2/blob/rel/2.17.2/log4j-layout-template-json/src/main/resources/GcpLayout.json).

One outstanding issue is converting the `trace_flags` hex value to a boolean, but it doesn't break Cloud Console if `logging.googleapis.com/trace_sampled` key is unset/false. See https://github.com/apache/logging-log4j2/discussions/2482.


examples/instrumentation-quickstart/src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,48 @@
...
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ -->
+
+<Configuration status="WARN">
+  <Appenders>
+    <Console name="Console" target="SYSTEM_OUT">
+      <!-- [START opentelemetry_instrumentation_setup_logging] -->
```
aabmass, line 20: This is a different region tag than the docs currently use https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/blob/97a8f20caf57263056ab0df8d8a54b4e7353eb2b/examples/instrumentation-quickstart/src/main/resources/logback.xml#L19

So I can update the docs before removing the logback.xml files

examples/instrumentation-quickstart/otel-collector-config.yaml
```diff
@@ -35,12 +35,26 @@ receivers:
...
-        field: body.trace_flags
-        value: "01"
-        if: body["logging.googleapis.com/trace_sampled"] == true
+      # TODO(aaronabbott): uncomment once the log4j2 config is correctly outputing the
+      # logging.googleapis.com/trace_sampled value
+      ## set trace_flags to SAMPLED if GCP attribute is set to true
+      # - type: add
+      #   field: body.trace_flags
+      #   value: "01"
+      #   if: body["logging.googleapis.com/trace_sampled"] == true
```
dashpole, line 57: if this returns false, will that just be a no-op? (do we actually need to comment out this config?)

dashpole (approved):

examples/instrumentation-quickstart/otel-collector-config.yaml
aabmass: Ah good point. I was trying to prevent collector log spam, but it looks like the log spam is actually coming from the `remove` operator:

`2024-04-18T14:52:50.317Z     error   helper/transformer.go:98        Failed to process entry {"kind": "receiver", "name": "filelog", "data_type": "logs", "operator_id": "remove6", "operator_type": "remove", "error": "remove: field does not exist: body['logging.googleapis.com/trace_sampled']", "action": "send", "entry": {"observed_timestamp":"2024-04-18T14:52:50.317546122Z","timestamp":"2024-04-18T14:52:50.214Z","body":{"_exception":{"stackTrace":""},"_logger":"com.example.demo.SingleController","_thread":"reactor-http-epoll-20","logging.googleapis.com/insertId":"2037","logging.googleapis.com/labels":{"span_id":"5aeb502bb6ef209f","trace_flags":"01","trace_id":"5643ea6d8d99b1b30a6fbab66b1a3ddd"},"logging.googleapis.com/sourceLocation":{"function":"com.example.demo.SingleController.handleSingle"},"message":"Going to sleep for 172"},"attributes":{"log.file.name":"app.log"},"severity_text":"INFO","span_id":"WutQK7bvIJ8=","trace_id":"VkPqbY2ZsbMKb7q2axo93Q==","severity":9,"scope_name":""}}`

I'll just let it be for now and reset this file

PR #333, opened 2024-04-19
Delete unused logback xml from instrumentation quickstart

aabmass:
This is no longer used since we switched the sample over to log4j2 in https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/pull/330

The snippet that was here is not used in the docs anymore: 
image


aabmass:
Just waiting for the usage of the old snippet to be removed from the internal doc sources

aabmass:
> Just waiting for the usage of the old snippet to be removed from the internal doc sources

It doesn't seem like it's going to happen. I think we are good to move forward with this PR

dashpole (approved):

PR #342, opened 2024-05-11
Fix #340 bug

tkmsaaaam:
Fix #340 's bug.

I made bellow changes.

## 1. Use appropriate `CLOUD_PROVIDER` . 
- `shared/resourcemapping/src/test/java/com/google/cloud/opentelemetry/resource/ResourceTranslatorTest.java` All or most of test case's  `CLOUD_PROVIDER` are `GCP`.  But I think it is not correct. It should map bellow.
  - When `GCP_KUBERNETES_ENGINE ` then `GCP`
  - When `AZURE_AKS` then `AZURE`
  - When `AWS_EKS` then `AWS`
  - When null then no key and value. (not cloud)
- This mapping follows Go's repository. Like here.  https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/683/files#diff-68a9a0e83deda8fe72338c49bcedd7da9b1633f67d40473059117d9db844529aR130-R343
- **So I fixed about `CLOUD_PROVIDER` in most of cases.**

## 2. Use appropriate `CLOUD_PLATFORM`.
- `shared/resourcemapping/src/test/java/com/google/cloud/opentelemetry/resource/ResourceTranslatorTest.java` All or most `non-cloud` test case's `CLOUD_PLATFORM` are `AZURE_AKS`. But it should be no keys and values.
  - here: https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/pull/342/files#diff-bbe578011b0d984210caf2aa4e41a2c6020f0580d15396d6be3be7aff4819289L223-L293
- **So I remove keys and values.**

## 3. Add test cases about `GCP` .
- I did it. https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/pull/342#discussion_r1597457360

## 4. Fix to map resource when `platform == null`

I have to map to k8s resources when this condition. If i did this.

> ## 2. Use appropriate `CLOUD_PLATFORM`.

Then I need duplicate these lines https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/pull/342/files#diff-992354fc4c9a5bcfac13812fb9e420d0498483ae56b986f25e71d6649386de99L187-L198 . So I make `mapK8sResourceOrGenericTaskOrNode` to refactor it.

### DIFF
--- a/shared/resourcemapping/src/main/java/com/google/cloud/opentelemetry/resource/ResourceTranslator.java
+++ b/shared/resourcemapping/src/main/java/com/google/cloud/opentelemetry/resource/ResourceTranslator.java
@@ -172,7 +172,7 @@ public class ResourceTranslator {
   public static GcpResource mapResource(Resource resource) {
     String platform = resource.getAttribute(ResourceAttributes.CLOUD_PLATFORM);
     if (platform == null) {
-      return genericTaskOrNode(resource);
+      return mapK8sResourceOrGenericTaskOrNode(resource);
     }
     switch (platform) {
       case ResourceAttributes.CloudPlatformValues.GCP_COMPUTE_ENGINE:
@@ -182,21 +182,25 @@ public class ResourceTranslator {
       case ResourceAttributes.CloudPlatformValues.GCP_APP_ENGINE:
         return mapBase(resource, "gae_instance", GOOGLE_CLOUD_APP_ENGINE_INSTANCE_LABELS);
       default:
-        // if k8s.cluster.name is set, pattern match for various k8s resources.
-        // this will also match non-cloud k8s platforms like minikube.
-        if (resource.getAttribute(ResourceAttributes.K8S_CLUSTER_NAME) != null) {
-          if (resource.getAttribute(ResourceAttributes.K8S_CONTAINER_NAME) != null) {
-            return mapBase(resource, "k8s_container", K8S_CONTAINER_LABELS);
-          } else if (resource.getAttribute(ResourceAttributes.K8S_POD_NAME) != null) {
-            return mapBase(resource, "k8s_pod", K8S_POD_LABELS);
-          } else if (resource.getAttribute(ResourceAttributes.K8S_NODE_NAME) != null) {
-            return mapBase(resource, "k8s_node", K8S_NODE_LABELS);
-          } else {
-            return mapBase(resource, "k8s_cluster", K8S_CLUSTER_LABELS);
-          }
-        }
-        return genericTaskOrNode(resource);
+        return mapK8sResourceOrGenericTaskOrNode(resource);
     }
+  }
+
+  private static GcpResource mapK8sResourceOrGenericTaskOrNode(Resource resource) {
+    // if k8s.cluster.name is set, pattern match for various k8s resources.
+    // this will also match non-cloud k8s platforms like minikube.
+    if (resource.getAttribute(ResourceAttributes.K8S_CLUSTER_NAME) != null) {
+      if (resource.getAttribute(ResourceAttributes.K8S_CONTAINER_NAME) != null) {
+        return mapBase(resource, "k8s_container", K8S_CONTAINER_LABELS);
+      } else if (resource.getAttribute(ResourceAttributes.K8S_POD_NAME) != null) {
+        return mapBase(resource, "k8s_pod", K8S_POD_LABELS);
+      } else if (resource.getAttribute(ResourceAttributes.K8S_NODE_NAME) != null) {
+        return mapBase(resource, "k8s_node", K8S_NODE_LABELS);
+      } else {
+        return mapBase(resource, "k8s_cluster", K8S_CLUSTER_LABELS);
+      }
+    }
+    return genericTaskOrNode(resource);
   }
 
   private static GcpResource genericTaskOrNode(Resource resource) {
--- a/shared/resourcemapping/src/test/java/com/google/cloud/opentelemetry/resource/ResourceTranslatorTest.java
+++ b/shared/resourcemapping/src/test/java/com/google/cloud/opentelemetry/resource/ResourceTranslatorTest.java
@@ -41,24 +41,8 @@ public class ResourceTranslatorTest {
 
   private static Stream<Arguments> provideOTelResourceAttributesToMonitoredResourceMapping() {
     return Stream.of(
-        generateOTelResourceMappingTestArgs(
-            Map.ofEntries(
-                new SimpleEntry<>(
-                    ResourceAttributes.CLOUD_PROVIDER, ResourceAttributes.CloudProviderValues.GCP),
-                new SimpleEntry<>(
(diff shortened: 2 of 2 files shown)


shared/resourcemapping/src/main/java/com/google/cloud/opentelemetry/resource/ResourceTranslator.java
```diff
@@ -182,21 +182,25 @@ public static GcpResource mapResource(Resource resource) {
...
-            return mapBase(resource, "k8s_container", K8S_CONTAINER_LABELS);
-          } else if (resource.getAttribute(ResourceAttributes.K8S_POD_NAME) != null) {
-            return mapBase(resource, "k8s_pod", K8S_POD_LABELS);
-          } else if (resource.getAttribute(ResourceAttributes.K8S_NODE_NAME) != null) {
-            return mapBase(resource, "k8s_node", K8S_NODE_LABELS);
-          } else {
-            return mapBase(resource, "k8s_cluster", K8S_CLUSTER_LABELS);
-          }
-        }
-        return genericTaskOrNode(resource);
```
psx95, line 198: If I'm reading this correctly, all of these lines are moved as-is to the `mapK8sResourceOrGenericTaskOrNode` function, correct ?

If this was indeed meant to be a refactor, I don't think we need it on account of my comment on Line 175.

shared/resourcemapping/src/main/java/com/google/cloud/opentelemetry/resource/ResourceTranslator.java
```diff
@@ -172,7 +172,7 @@ public static AttributeMapping create(
   public static GcpResource mapResource(Resource resource) {
     String platform = resource.getAttribute(ResourceAttributes.CLOUD_PLATFORM);
     if (platform == null) {
-      return genericTaskOrNode(resource);
+      return mapK8sResourceOrGenericTaskOrNode(resource);
```
psx95, line 175: Could you explain your reasoning behind this? 

If platform is `null`, we do not want it mapped to K8s.

shared/resourcemapping/src/test/java/com/google/cloud/opentelemetry/resource/ResourceTranslatorTest.java
```diff
@@ -88,6 +88,60 @@ private static Stream<Arguments> provideOTelResourceAttributesToMonitoredResourc
             Map.ofEntries(
                 new SimpleEntry<>(
                     ResourceAttributes.CLOUD_PROVIDER, ResourceAttributes.CloudProviderValues.GCP),
+                new SimpleEntry<>(
```
psx95, line 91: Just confirming here: The newly added tests here are test cases that contain GCP Platform values, correct ?

psx95 (changes requested):

psx95:
> Fix #340 's bug.
> 
> * Use appropriate `CLOUD_PROVIDER` .
> * Add test cases about `GCP` .
> * Fix to map resource when `platform == null`

 - Regarding use of appropriate cloud provider, I think GCP is correct, since [GCP has Monitored Resource types for other cloud providers like AWS too](https://cloud.google.com/monitoring/api/resources#tag_aws_ec2_instance).

shared/resourcemapping/src/main/java/com/google/cloud/opentelemetry/resource/ResourceTranslator.java
tkmsaaaam: It is correct.

shared/resourcemapping/src/test/java/com/google/cloud/opentelemetry/resource/ResourceTranslatorTest.java
tkmsaaaam: It is correct.

tkmsaaaam:
I update description. https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/pull/342#issue-2290726664

tkmsaaaam:
> Fix https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/pull/340 's bug.
> 
> Use appropriate CLOUD_PROVIDER .
> Add test cases about GCP .
> Fix to map resource when platform == null
> Regarding use of appropriate cloud provider, I think GCP is correct, since [GCP has Monitored Resource types for other cloud providers like AWS too](https://cloud.google.com/monitoring/api/resources#tag_aws_ec2_instance).

If it is true. I don't need changes about below. I will fix it.

> ## 1. Use appropriate CLOUD_PROVIDER .

psx95:
Thank you for the clarification, I will review this shortly!

shared/resourcemapping/src/test/java/com/google/cloud/opentelemetry/resource/ResourceTranslatorTest.java
psx95, line 1: Since the number of test arguments have increased a lot in this file, could you also "group" them by `expectedMRType` ? For instance - 
```java
 return Stream.of(
        // Test cases for MR type gce_instance
        generateOTelResourceMappingTestArgs(..., "gce_instance", ...),
        ...,
        // Test cases for MR type k8s_container
         generateOTelResourceMappingTestArgs(..., "k8s_container", ...),
         generateOTelResourceMappingTestArgs(..., "k8s_container", ...),
        ...,
        // Test cases MR type k8s_node
        generateOTelResourceMappingTestArgs(..., "k8s_node", ...),
        ...
```
This would help in keeping test cases maintainable.

psx95 (changes requested):
Thanks for reporting and fixing this bug!
 Could you address the last [pending comment](https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/pull/342/files#r1600598037)

tkmsaaaam:
> Thanks for reporting and fixing this bug!
> Could you address the last [pending comment](https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/pull/342/files#r1600598037)

Thanks for your review. I've done it.

psx95 (approved):
/gcbrun

psx95:
> > Fix #340 's bug.
> > Use appropriate CLOUD_PROVIDER .
> > Add test cases about GCP .
> > Fix to map resource when platform == null
> > Regarding use of appropriate cloud provider, I think GCP is correct, since [GCP has Monitored Resource types for other cloud providers like AWS too](https://cloud.google.com/monitoring/api/resources#tag_aws_ec2_instance).
> 
> If it is true. I don't need changes about below. I will fix it.
> 
> > ## 1. Use appropriate CLOUD_PROVIDER .

For posterity, the `CLOUD_PROVIDER` resource attribute is not added to the MR resource, as such its value does not affect the mapped resource. However, the more appropriate `CLOUD_PROVIDER` values should be specific to the `CLOUD_PLATFORM` as indicated by the tests in this PR.

PR #346, opened 2024-05-19
Add support for custom MR mappings

psx95:
### Description 
Provides a configuration option for the metrics exporter allowing users to specify mappings for custom MonitoredResource type. 

The mapping logic for the custom MR type is activated based on the presence of `gcp.resource_type` resource attribute in the underlying OpenTelemetry resource.

Some methods were marked with `@Deprecated` but no breaking changes were made in this PR. 

#### Sample usage with the proposed changes

```java
  // MetricConfiguration object for the metrics exporter - 
  MetricConfiguration configuration =
        MetricConfiguration.builder()
            .setProjectId(PROJECT_ID)
            .setCredentials(CREDENTIALS)
            .setResourceAttributesFilter(allowAllPredicate)
            .setMonitoredResourceDescription(
                new MonitoredResourceDescription(
                    "custom_mr", Set.of("instance_id", "api", "host_id")))
            .setUseServiceTimeSeries(true)
            .build();
            
  // If the OTel resource now has the Attribute label `gcp.resource_type` with value 
  // `custom_mr` the exporter will attempt to map the OTel resource to the set MR type.
```

### Testing
 - Tested this newly added configuration with the existing metrics example.
 - Used the configuration to override MR type to dataflow-job.
 - The program returned valid error when all required labels were not present in the resource attributes.
 - When all required labels were present, metric was successfully written against the correct MR type.

###### Configuration code for reference
```java

// MeterProvider and OTel Resource setup
    MonitoredResourceDescription monitoredResourceDescription =
        new MonitoredResourceDescription(
            "dataflow_job", Set.of("job_name", "project_id", "region"));

    MetricExporter metricExporter =
        GoogleCloudMetricExporter.createWithConfiguration(
            MetricConfiguration.builder()
                .setMonitoredResourceDescription(monitoredResourceDescription)
                .build());

    // project_id inferred from env variables
    Attributes attrs =
        Attributes.builder()
            .put("gcp.resource_type", "dataflow_job") // required to trigger platform mapping
            .put("network_id", "random-id") // was ignored
            .put("job_name", "some-job")
            .put("region", "us1")
            .build();

    METER_PROVIDER =
        SdkMeterProvider.builder()
            .setResource(Resource.create(attrs))
            .registerMetricReader(
                PeriodicMetricReader.builder(metricExporter)
                    .setInterval(java.time.Duration.ofSeconds(30))
                    .build())
            .build();
  ...
  // Standard OpenTelemetry code to record metric
```

### DIFF
--- a/exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/MetricConfiguration.java
+++ b/exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/MetricConfiguration.java
@@ -46,7 +46,7 @@ public abstract class MetricConfiguration {
   /** Resource attribute filter that disables addition of resource attributes to metric labels. */
   public static final Predicate<AttributeKey<?>> NO_RESOURCE_ATTRIBUTES = attributeKey -> false;
 
-  public static final MonitoredResourceDescription DEFAULT_MONITORED_RESOURCE_DESCRIPTION =
+  public static final MonitoredResourceDescription EMPTY_MONITORED_RESOURCE_DESCRIPTION =
       new MonitoredResourceDescription("", Collections.emptySet());
 
   /**
@@ -160,8 +160,8 @@ public abstract class MetricConfiguration {
    * {@link io.opentelemetry.sdk.resources.Resource} to Google specific {@link
    * com.google.api.MonitoredResource}.
    *
-   * <p>This returns the {@link MetricConfiguration#DEFAULT_MONITORED_RESOURCE_DESCRIPTION} if not
-   * set through exporter configuration.
+   * <p>This returns the {@link MetricConfiguration#EMPTY_MONITORED_RESOURCE_DESCRIPTION} if not set
+   * through exporter configuration.
    *
    * @return The {@link MonitoredResourceDescription} object containing the MonitoredResource type
    *     and its expected labels.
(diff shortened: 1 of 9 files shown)


exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/MetricConfiguration.java
```diff
@@ -45,6 +46,9 @@ public abstract class MetricConfiguration {
   /** Resource attribute filter that disables addition of resource attributes to metric labels. */
   public static final Predicate<AttributeKey<?>> NO_RESOURCE_ATTRIBUTES = attributeKey -> false;
 
+  public static final MonitoredResourceDescription DEFAULT_MONITORED_RESOURCE_DESCRIPTION =
```
dashpole, line 49: nit: should we call this "empty monitored resource description"?  It is also the default, but its probably more important that it is empty or the zero state.

exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/ResourceTranslator.java
```diff
@@ -17,13 +17,24 @@
...
+import io.opentelemetry.api.common.AttributeKey;
 import io.opentelemetry.sdk.resources.Resource;
+import java.util.Objects;
+import java.util.Set;
+import java.util.logging.Logger;
 
 /** Translates from OpenTelemetry Resource into Google Cloud Monitoring's MonitoredResource. */
 public class ResourceTranslator {
+  private static final String CUSTOM_MR_KEY = "gcp.resource_type";
+  private static final String BLANK_STRING = " ";
```
dashpole, line 30: Should this be the empty string, or does it need to have a space?

dashpole (approved):
lgtm after comment is addressed

exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/ResourceTranslator.java
psx95: Discussed internally, this will not be required - only sending labels that are found and relying on back-end to reject malformed MR.

jsuereth (approved):

PR #350, opened 2024-05-26
Make the trace flags optional when parsing x-cloud-trace-context

sawadakaku:
Hello guys,

I've observed that Cloud load balancing sends requests with `x-cloud-trace-context` that lack `o=<sampled>` to the backend service.
This is causing the propagator not properly to propagate the trace context.
I addressed this issue by making the trace flag optional.

Could someone look over this?

Ref: https://github.com/GoogleCloudPlatform/opentelemetry-operations-python/issues/167


psx95 (approved):

psx95:
Fixes #351

PR #353, opened 2024-06-22
Add Java8 support

psx95:


build.gradle
```diff
@@ -234,12 +240,6 @@ subprojects {
 		archives javadocJar, sourcesJar
 	}
 
-	javadoc {
```
psx95, line 237: Minimum JDK version required for building this project is JDK 11, html5 is default.

build.gradle
```diff
@@ -103,6 +100,26 @@ subprojects {
...
+			languageVersion = JavaLanguageVersion.of(11)
+		}
+	}
+
+	// This ensures bytecode compatibility with Java 8 by ensuring that symbols not
+	// available in Java 8 are not used in the source code.
+	// This is equivalent of setting --release flag in Java compiler introduced in Java 9.
+	// Note that the toolchain used is Java 11 - which means Java 11 is required
+	// to build this project.
+	afterEvaluate {
```
aabmass, line 117: nit does this need to be in `afterEvaluate`? https://github.com/search?type=code&q=options.release+%3D+8+lang%3Agradle

e2e-test-server/build.gradle
```diff
@@ -23,6 +23,16 @@ application {
 	mainClass.set('com.google.cloud.opentelemetry.endtoend.Server')
 }
 
+// Java 11 required to build this module since it has a dependency on Cloud Functions.
+// Cloud Functions (Gen2) have Java 11 minimum requirement.
+afterEvaluate {
```
aabmass, line 28: same comment as above

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/InternalTraceExporter.java
```diff
@@ -53,7 +56,10 @@ class InternalTraceExporter implements SpanExporter {
   private final TraceTranslator translator;
 
   private static final Map<String, String> HEADERS =
-      Map.of("User-Agent", "opentelemetry-operations-java/" + TraceVersions.EXPORTER_VERSION);
+      Stream.of(
```
aabmass, line 59: Could use `singletonMap()` here? https://docs.oracle.com/javase/8/docs/api/java/util/Collections.html#singletonMap-K-V-

aabmass (approved):
@punya any concerns about not actually testing against Java8? @psx95 mentioned that we weren't doing it before either so this seems reasonable. But I'm no Java expert and don't know what might go wrong

exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/InternalTraceExporter.java
psx95: Good catch, yes - I could replace it.

build.gradle
psx95: No, it looks like `afterEvaluate` is extraneous here.

psx95:
> @punya any concerns about not actually testing against Java8? @psx95 mentioned that we weren't doing it before either so this seems reasonable. But I'm no Java expert and don't know what might go wrong

I think the ideal way going forward would be to write a CI that loads the modules/JAR intended to be distributed as Java8 compatible on a VM that only has Java 8 installed on it. 

I don't think we should force the current unit tests to use Java8 API.

Issue #149, opened 2022-04-29
Getting useful errors when metrics fail - Header size exceeded max allowed size


pulewicz:
Hi,

We've added google metrics exporter to our codebase. It works locally, however when we deploy it on GKE, we get the following error (full stacktrace at the end of the post):

```
io.grpc.netty.shaded.io.netty.handler.codec.http2.Http2Exception: Header size exceeded max allowed size (10240)
```

I think what's happening is we get some HTTP error response, but we can't decode its because of the header size limit. Is it possible to increase the limit or get more information about this error some other way?

Full stacktrace:

```
[otel.javaagent 2022-04-28 18:05:34:341 +0000] [PeriodicMetricReader-1] WARN io.opentelemetry.sdk.metrics.export.PeriodicMetricReader - Exporter threw an Exception
com.google.api.gax.rpc.InternalException: io.grpc.StatusRuntimeException: INTERNAL: http2 exception
	at com.google.api.gax.rpc.ApiExceptionFactory.createException(ApiExceptionFactory.java:67)
	at com.google.api.gax.grpc.GrpcApiExceptionFactory.create(GrpcApiExceptionFactory.java:72)
	at com.google.api.gax.grpc.GrpcApiExceptionFactory.create(GrpcApiExceptionFactory.java:60)
	at com.google.api.gax.grpc.GrpcExceptionCallable$ExceptionTransformingFuture.onFailure(GrpcExceptionCallable.java:97)
	at com.google.api.core.ApiFutures$1.onFailure(ApiFutures.java:68)
	at com.google.common.util.concurrent.Futures$CallbackListener.run(Futures.java:1074)
	at com.google.common.util.concurrent.DirectExecutor.execute(DirectExecutor.java:30)
	at com.google.common.util.concurrent.AbstractFuture.executeListener(AbstractFuture.java:1213)
	at com.google.common.util.concurrent.AbstractFuture.complete(AbstractFuture.java:983)
	at com.google.common.util.concurrent.AbstractFuture.setException(AbstractFuture.java:771)
	at io.grpc.stub.ClientCalls$GrpcFuture.setException(ClientCalls.java:563)
	at io.grpc.stub.ClientCalls$UnaryStreamToFuture.onClose(ClientCalls.java:533)
	at io.grpc.internal.ClientCallImpl.closeObserver(ClientCallImpl.java:553)
[...]

PR #356, opened 2024-07-04
Expose option to provide MetricServiceSettings in MetricConfiguration

psx95:
#### Description
This PR adds a configuration option to the `MetricConfiguration` that allows the user to configure the underlying instance of `MetricServiceClient` which is used by the exporter to interact with the Cloud Monitoring API.

This feature is intended for advanced use-cases only.

#### Testing
 - Added unit tests pass.
 - Updated the metrics sample to demonstrate a use-case for the `setMetricServiceSettings` function.

fixes. #149

### DIFF
--- a/exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/MetricConfiguration.java
+++ b/exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/MetricConfiguration.java
@@ -20,6 +20,7 @@ import static java.time.Duration.ZERO;
 import com.google.auth.Credentials;
 import com.google.auto.value.AutoValue;
 import com.google.cloud.ServiceOptions;
+import com.google.cloud.monitoring.v3.MetricServiceSettings;
 import com.google.cloud.monitoring.v3.stub.MetricServiceStubSettings;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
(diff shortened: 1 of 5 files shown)


exporters/metrics/src/main/java/com/google/cloud/opentelemetry/metric/MetricConfiguration.java
```diff
@@ -282,6 +296,29 @@ public abstract Builder setMonitoredResourceDescription(
...
+    /**
+     * Sets the options used to configure the {@link
+     * com.google.cloud.monitoring.v3.MetricServiceClient} used to interact with the Cloud
+     * Monitoring API. This is for advanced usage and must be configured carefully.
+     *
+     * <p>Providing MetricServiceSettings will cause the exporter to ignore the values configured
+     * using:
+     *
+     * <ul>
+     *   <li>{@link MetricConfiguration.Builder#setInsecureEndpoint(boolean)}
```
dashpole, line 308: Did you explore making these functions work with or update MetricServiceSettings?

psx95: Yes, I did try to reconcile these functions with the `setMetricServiceSettings`, but it lead to a confusing API. All of these functions eventually affect the properties of the underlying `MetricServiceSettings` object used to create the client.
 
So in a config like:
```java
    return MetricConfiguration.builder()
        .setMetricServiceSettings(metricServiceSettingsBuilder.build())
        .setCredentials(validCredentials)
        // sets the channel to use NoCredentialsProvider and plainText.
        .setInsecureEndpoint(true)
        .setMetricServiceEndpoint(someEndpoint)
        .build();
```

It would become confusing for the user to predict/understand what the final configuration holds. Ignoring these other settings when `setMetricServiceSettings` is used keeps the **implementation clean** and the intent of the API clear to the user.

We also have exposed the defaults which the users can plug-in when constructing the `MetricServiceSettings` object.

dashpole (approved):

PR #364, opened 2024-08-23
Update the collector image used in instrumentation quickstart

aabmass:
Updated the otelcol image and fixed substitution syntax `${}` issue

Since the gRPC metrics are no longer available, tests were broken. I chose to rewrite the integration test in Go so we can easily use it in the other 3 samples.


dashpole (approved):

aabmass:
Tests are broken because of https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/797. I can use a different metric instead but will need to modify the tests a little bit. I'm thinking to rewrite in Go so we can use the same integration test setup across all the repos

examples/instrumentation-quickstart/integration_test/instrumentation_test.go
```diff
@@ -0,0 +1,154 @@
...
+		return "", err
+	}
+	collectorPort, err := collectorContainer.MappedPort(ctx, "8888")
+	if err != nil {
+		return "", err
+	}
+	return fmt.Sprintf("http://%s:%s/metrics", collectorHost, collectorPort.Port()), nil
+}
+
+func verifyPromMetric(t assert.TestingT, promMetrics map[string]*dto.MetricFamily, tc testCase) {
```
aabmass, line 138: @dashpole is there a better way to make assertions against prometheus text format? These seem to be protobuf structs which are not very ergonomic and possibly deprecated. I could do regex like we did for Java, but seems a little fragile.

examples/instrumentation-quickstart/integration_test/instrumentation_test.go
aabmass, line 1: This test should work identically across the different samples we have.

- Is there an easy way to share this test across repos? I think we could do `go install thetest@latest && go test thetest`. Or refactor this into a lib and leave a small go test in each repo.
- Where should the code live?

examples/instrumentation-quickstart/integration_test/instrumentation_test.go
dashpole: Consider using [testutil.ScrapeAndCompare](https://github.com/prometheus/client_golang/blob/97aa0493ebfa75f20fe51ebb32f1e5c5315e5fa1/prometheus/testutil/testutil.go#L162).  Its usage is similar to GatherAndCompare, which we use to test the otel-go prometheus exporter: https://github.com/open-telemetry/opentelemetry-go/blob/48fedfa58e35b5f3b08fe54e45cfad441dd82426/exporters/prometheus/exporter_test.go#L449

examples/instrumentation-quickstart/integration_test/instrumentation_test.go
dashpole: I would consider putting it in opentelemetry-operations-e2e-testing or opentelemetry-operations-go.

Any of those options sound reasonable.  If it was in opentelemetry-operations-e2e-testing, it would be easy to publish as a docker container along with the other test image.  Otherwise, I would lean towards refactoring it into a library and putting it in one of those repos.

examples/instrumentation-quickstart/integration_test/instrumentation_test.go
aabmass: It looks like this requires exact numeric values, so I'm not sure I can get it working in an un-flaky way. I'll look through the other testutil though, thanks!

examples/instrumentation-quickstart/integration_test/instrumentation_test.go
aabmass: Agreed, moved to https://github.com/GoogleCloudPlatform/opentelemetry-operations-e2e-testing/pull/58 as a small test lib.

PR #384, opened 2024-10-29
Hardcode versions on all samples

psx95:
This PR hard-codes version number on all samples since they are not meant to be published on maven. 

Hardcoding samples simplifies writing docs and instructions as it gives us deterministic file names from the build output.

Also removes `group` declaration on a few example modules to keep things consistent.

### DIFF
--- a/examples/autoconf/build.gradle
+++ b/examples/autoconf/build.gradle
@@ -19,8 +19,10 @@ plugins {
 	id 'com.google.cloud.tools.jib'
 }
 
-
 description = 'Examples for using java autoconfiguration and Google Cloud Operations'
+
+// examples are not published, so version can be hardcoded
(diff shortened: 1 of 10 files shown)


examples/autoconf/build.gradle
```diff
@@ -19,9 +19,11 @@ plugins {
 	id 'com.google.cloud.tools.jib'
 }
 
-
 description = 'Examples for using java autoconfiguration and Google Cloud Operations'
 
+// examples are not published, so version can be hardcoded
+version = '1.0.0'
```
dashpole, line 25: can we use 0.1.0 to ensure these aren't be seen as stable, even if they aren't published?

punya (approved):

PR #388, opened 2025-01-14
Add otlp metric example with gcp auth

dashpole:
This is a copy of https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/tree/main/examples/otlptrace, but changed to use metrics (based on https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/blob/main/examples/metrics/src/main/java/com/google/cloud/opentelemetry/example/metrics/MetricsExporterExample.java).

TODO: test


dashpole:
@psx95 can you test this for me? (just make sure it builds)

psx95 (approved):

PR #391, opened 2025-01-31
Update OTLP Trace Example to use Auth Extension

psx95:
Updates the OTLP Trace example to use the [GCP Auth Extension](https://github.com/open-telemetry/opentelemetry-java-contrib/tree/main/gcp-auth-extension) to export traces to GCP.

### DIFF
--- a/examples/otlptrace/build.gradle
+++ b/examples/otlptrace/build.gradle
@@ -27,17 +27,16 @@ dependencies {
 	implementation(libraries.opentelemetry_sdk)
 	implementation(libraries.opentelemetry_otlp_exporter)
 	implementation(libraries.opentelemetry_sdk_autoconf)
-	implementation(libraries.google_auth)
-	implementation(libraries.opentelemetry_gcp_auth_extension)
+	// the shaded variant name is empty
+	implementation("${libraries.opentelemetry_gcp_auth_extension}:")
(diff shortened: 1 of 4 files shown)


dashpole (approved):

examples/otlptrace/build.gradle
```diff
@@ -20,28 +20,31 @@ plugins {
...
 description = 'Example for OTLP exporter'
 
 dependencies {
 	implementation(libraries.opentelemetry_api)
 	implementation(libraries.opentelemetry_sdk)
 	implementation(libraries.opentelemetry_otlp_exporter)
 	implementation(libraries.opentelemetry_sdk_autoconf)
-	implementation(libraries.google_auth)
+	// the shaded variant name is empty
+	implementation("${libraries.opentelemetry_gcp_auth_extension}:")
```
aabmass, line 31: 🙈 

Not sure if it's only .kts, but is there a way to pass named arguments so you could make the empty string more explicit?

psx95: Yeah for some reason,
```groovy
implementation group: 'io.opentelemetry.contrib', name: 'opentelemetry-gcp-auth-extension', version: '1.43.0-alpha', classifier: ''
```
is not working. 

I made a change that makes the intentional use of empty classifier more explicit. This is temporary and will be fixed in the next release of the auth extension.

PR #402, opened 2025-03-17
Fix the error in releasing instructions

psx95:
Updates the release instructions to match the current setup.

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -52,8 +52,8 @@ you can configure the following in `<your-home-directory>/.gradle/gradle.propert
     ```text
     ossrhUsername=<generated-token-user>
     ossrhPassword=<generated-token-key>
-    signingUseGpgCmd=true
 
+    signingUseGpgCmd=true
     signing.gnupg.executable=gpg
     signing.gnupg.keyName=<secret key id (large hash)>
     ```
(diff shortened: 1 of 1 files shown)


RELEASING.md
```diff
@@ -50,15 +50,18 @@ If you're running in linux and would like to use the GPG agent to remember your
...
+Note: You can retrieve the list of previously created GPG keys on your machine by using `gpg --list-secret-keys`.
 
 > [!IMPORTANT]
 > Starting June 2024, due to a change to the OSSRH authentication backend, the maven publish plugin now requires [a user token](https://central.sonatype.org/publish/generate-token/) instead of a typical username and password used in the Nexus UI.
-> Follow the steps in the [link](https://central.sonatype.org/publish/generate-token/) to generate a user token, if not done already - this will provide you with a `tokenuser` and `tokenkey`. Replace the `ossrhUsername` and `ossrhPassword` with this `tokenuser` and `tokenkey` in your `gradle.properties` file to successfully publish artifacts.
+> Follow the steps in the [link](https://central.sonatype.org/publish/generate-token/) to generate a user token, if not done already - this will provide you with a `tokenuser` and `tokenkey`. Replace the fields `ossrhUsername` and `ossrhPassword` with the values for `tokenuser` and `tokenkey` in your `gradle.properties` file to successfully publish artifacts.
```
quentinmit, line 64: How about

```
Replace `<generated-token-user>` and `<generated-token-key>` with the token and key you generated.
```

RELEASING.md
```diff
@@ -50,15 +50,18 @@ If you're running in linux and would like to use the GPG agent to remember your
 you can configure the following in `<your-home-directory>/.gradle/gradle.properties`:
 
     ```text
+    ossrhUsername=<generated-token-user>
+    ossrhPassword=<generated-token-key>
+    signingUseGpgCmd=true
+
```
quentinmit, line 56: Nit: This whitespace seems unnecessary (and if you do want to split them into two groups, it seems like `signingUseGpgCmd` belongs in the second?)

quentinmit (approved):

PR #410, opened 2025-05-15
Update otlpmetrics sample

psx95:
#### Description
Updates the OTLP sample with the following changes: 
1. Defers adding quota project header to Google auth library.
2. Adds GCP resource detection.
3. Updates documentation.
4. Shows how to use dynamic headers to enable auto token refresh.

#### Testing
 - Manually ran sample with End User Credentials and Service Credentials. 
 - Confirmed metrics exported on GCP console.

### DIFF
--- a/examples/otlpmetric/src/main/java/com/google/cloud/opentelemetry/example/otlpmetric/OTLPMetricExample.java
+++ b/examples/otlpmetric/src/main/java/com/google/cloud/opentelemetry/example/otlpmetric/OTLPMetricExample.java
@@ -62,6 +62,9 @@ public class OTLPMetricExample {
                     .addHeader(
                         "Authorization", "Bearer " + credentials.getAccessToken().getTokenValue());
 
+        if (credentials.getQuotaProjectId() != null) {
+          builder.addHeader("x-goog-user-project", credentials.getQuotaProjectId());
+        }
         return builder.build();
       } catch (IOException e) {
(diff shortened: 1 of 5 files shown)


examples/otlpmetric/src/main/java/com/google/cloud/opentelemetry/example/otlpmetric/OTLPMetricExample.java
```diff
@@ -74,6 +77,10 @@ private static MetricExporter addAuthorizationHeaders(
                 .toBuilder()
                     .addHeader(
                         "Authorization", "Bearer " + credentials.getAccessToken().getTokenValue());
+
+        if (credentials.getQuotaProjectId() != null) {
```
aabmass, line 81: Does the Authorization header work with refresh as is?

psx95: No, the refresh will not work in this sample yet - this sample still needs to be updated to use the auth extension which would handle this.
Additionally, the sample in itself is correct - the use case shown here does not require refresh.

This change is made to reduce confusion around the quota headers and adding GCP resource detection. I'm working on adding metrics support in the auth extension, after which this sample would require another update.

aabmass (approved):

PR #435, opened 2026-04-01
Add autoinstrument example with OTLP endpoints

psx95:


examples/autoinstrument-auth-extension/README.md
```diff
@@ -0,0 +1,110 @@
...
+##### Get Google Cloud Credentials on your machine
+
+```shell
+gcloud auth application-default login
+```
+Executing this command will save your application credentials to default path which will depend on the type of machine -
+- Linux, macOS: `$HOME/.config/gcloud/application_default_credentials.json`
+- Windows: `%APPDATA%\gcloud\application_default_credentials.json`
+
+**NOTE: This method of authentication is not recommended for production environments.**
```
dashpole: Should we link to something, or say what users should do?

dashpole (approved):

examples/autoinstrument-auth-extension/README.md
psx95: Yeah, could be a good idea. Do you have the instructions handy ?
Otherwise I can take an AI to add them later.

dashpole: Maybe https://docs.cloud.google.com/docs/authentication#auth-decision-tree?

PR #441, opened 2026-05-07
Deprecate the exporters and related artifacts

psx95:
This PR marks the public APIs of the artifacts released from this repository as deprecated and also emits deprecation warnings during runtime in form of logs in case any of these artifacts are used in an application.

The PR also adds a migration guide, which will hold instructions for users to follow while moving from the OpenTelemetry Google Cloud Monitoring and Trace exporters to OpenTelemetry OTLP exporters.

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -21,8 +21,6 @@ Grant the following IAM roles to the service account that your application uses:
 
 *   **Cloud Telemetry Traces Writer** (`roles/telemetry.tracesWriter`)
 *   **Cloud Telemetry Metrics Writer** (`roles/telemetry.metricsWriter`)
-*   **Logs Writer** (`roles/logging.logWriter`)
-*   **Monitoring Metric Writer** (`roles/monitoring.metricWriter`)
 
 ## Migrate from OpenTelemetry Google Cloud Trace Exporter to OTLP exporter
 
@@ -203,9 +201,9 @@ The following table maps the configurations available in `MetricConfiguration` t
 The following features of the `GoogleCloudMetricExporter` are not supported by the standard OTLP exporter:
 
 *   **Metric Descriptor Strategy (`setDescriptorStrategy`)**: OTLP exporters do not send metric descriptors separately. Metadata is handled automatically by the backend.
-*   **Custom Monitored Resource Mapping (`setMonitoredResourceDescription`)**: OTLP relies on standard OTel resources. GCP maps these to monitored resources automatically.
+*   **Custom Monitored Resource Mapping (`setMonitoredResourceDescription`)**: OTLP relies on standard OTel resources. GCP maps these to monitored resources automatically. This feature was added to support internal use-cases only.
 *   **Predicate-based Resource Attribute Filtering (`setResourceAttributesFilter`)**: OTLP exporters send all resource attributes by default. If you need to filter them, you must do so before they reach the exporter (e.g., via resource configuration or a processor if using a collector).
-*   **Use Service Time Series (`setUseServiceTimeSeries`)**: This option is specific to the Cloud Monitoring API and is not available in OTLP exporters.
+*   **Use Service Time Series (`setUseServiceTimeSeries`)**: This option is specific to the Cloud Monitoring API and is not available in OTLP exporters. This feature is for supporting internal use-cases only.
 *   **Instrumentation Library Labels Toggle (`setInstrumentationLibraryLabelsEnabled`)**: OTLP exporters send instrumentation scope information by default. Disabling it requires dropping the attributes via views or processors.
 *   **Custom Metric Service Settings (`setMetricServiceSettings`)**: You cannot pass `MetricServiceSettings` to OTLP exporters. If you need custom channel or client configuration, you must use programmatic configuration with `OtlpGrpcMetricExporter.builder()` or `OtlpHttpMetricExporter.builder()`.
 
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -9,7 +9,7 @@ Google Cloud now supports native OTLP (OpenTelemetry Protocol) ingestion for Clo
 
 To migrate from the deprecated Google Cloud Trace exporter to the standard OpenTelemetry OTLP exporter, follow these steps:
 
-### 1. Add Dependencies
+#### 1. Add Dependencies
 
 Add the following dependencies to your `build.gradle` file:
 
@@ -20,7 +20,7 @@ implementation("io.opentelemetry:opentelemetry-sdk-extension-autoconfigure:1.56.
 implementation("io.opentelemetry.contrib:opentelemetry-gcp-auth-extension:1.52.0-alpha")
 ```
 
-### 2. Configure the SDK
+#### 2. Configure the SDK
 
 Use the OpenTelemetry SDK Autoconfigure module to configure the SDK. You can set the following system properties or environment variables:
 
@@ -36,7 +36,13 @@ OTEL_TRACES_EXPORTER=otlp
 OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf
 ```
 
-### Configuration Mapping
+#### 3. Follow the Migration Guide
+
+For a code walkthrough, follow the migration guide published at [Migrate from the Trace exporter to the OTLP endpoint](https://docs.cloud.google.com/trace/docs/migrate-to-otlp-endpoints).
+
+### Mapping and Limitations 
+
+#### Configuration Mapping
 
 The following table maps the configurations available in `TraceConfiguration` to their OTLP equivalents:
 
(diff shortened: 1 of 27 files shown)


MIGRATION.md
```diff
@@ -0,0 +1,22 @@
+# Migration Guide
+
+This guide provides instructions on how to migrate from the custom exporters in this repository to the standard OpenTelemetry OTLP exporters.
+
+## Overview
+Google Cloud now supports native OTLP (OpenTelemetry Protocol) ingestion for Cloud Trace and Cloud Monitoring via the [Telemetry API](https://docs.cloud.google.com/stackdriver/docs/reference/telemetry/overview). This allows you to use the standard OpenTelemetry OTLP exporters for sending telemetry data to Google Cloud.
+
+## Migrate from OpenTelemetry Google Cloud Trace Exporter to OTLP exporter
+
+Follow the [Migrate from the Trace exporter to the OTLP endpoint](https://docs.cloud.google.com/trace/docs/migrate-to-otlp-endpoints) guide for migration instructions.
```
dashpole: I think you can use https://docs.cloud.google.com/trace/docs/migrate-to-otlp-endpoints#java_3 to make sure it shows the java samples?

MIGRATION.md
```diff
@@ -0,0 +1,22 @@
+# Migration Guide
+
+This guide provides instructions on how to migrate from the custom exporters in this repository to the standard OpenTelemetry OTLP exporters.
+
+## Overview
+Google Cloud now supports native OTLP (OpenTelemetry Protocol) ingestion for Cloud Trace and Cloud Monitoring via the [Telemetry API](https://docs.cloud.google.com/stackdriver/docs/reference/telemetry/overview). This allows you to use the standard OpenTelemetry OTLP exporters for sending telemetry data to Google Cloud.
+
+## Migrate from OpenTelemetry Google Cloud Trace Exporter to OTLP exporter
```
dashpole, line 8: I was imagining this would be fairly detailed, but I guess it doesn't have to be if it is covered in the GCP docs. We should make sure we tell users of each of the configuration options how they can obtain the same behavior with the OTLP exporter:

* setFixedAttributes
* setAttributeMapping

MIGRATION.md
```diff
@@ -0,0 +1,22 @@
...
+
+Follow the [Migrate from the Trace exporter to the OTLP endpoint](https://docs.cloud.google.com/trace/docs/migrate-to-otlp-endpoints) guide for migration instructions.
+
+## Migrate from OpenTelemetry Google Cloud Monitoring Exporter to OTLP exporter
+
+> [!NOTE] The Google Cloud OTLP metrics endpoint is currently in preview and the migration guides are being developed. 
+
+TODO: Add migration guide for metrics exporter
```
dashpole: Things to cover for metrics other than just the default "make it work":

* How to set a prefix with the OTLP exporter (Currently using setPrefix, or if they want the previous default workload.googleapis.com)
* setResourceAttributesFilter
* setInstrumentationLibraryLabelsEnabled

MIGRATION.md
psx95: If I replace it with this, it would directly move to the code block skipping the pre-requisites like configuring permissions, etc. 
I think this link is ok. Additionally, Java is also the the first listed language - so its selected by default.

MIGRATION.md
psx95: I have updated the doc and added the migration steps.

MIGRATION.md
```diff
@@ -0,0 +1,220 @@
...
+2.  Verify that billing is enabled for your Google Cloud project.
+3.  Enable the Telemetry, Cloud Logging, Cloud Monitoring, and Cloud Trace APIs.
+
+### Configure permissions
+
+Grant the following IAM roles to the service account that your application uses:
+
+*   **Cloud Telemetry Traces Writer** (`roles/telemetry.tracesWriter`)
+*   **Cloud Telemetry Metrics Writer** (`roles/telemetry.metricsWriter`)
+*   **Logs Writer** (`roles/logging.logWriter`)
```
dashpole: Not sure if this is relevant for the java migration

MIGRATION.md
```diff
@@ -0,0 +1,220 @@
...
+| `setDeadline(Duration)` | `otel.exporter.otlp.timeout` / `OTEL_EXPORTER_OTLP_TIMEOUT` | Default is 10 seconds. |
+| `setPrefix(String)` | N/A | The Telemetry API automatically prefixes metrics with `prometheus.googleapis.com/` by default and will be stored in Google Managed Prometheus with `prometheus_target` as the Monitored Resource. Custom prefixes are not directly supported via OTLP exporter configuration.<br>If you want your metric to keep appearing under your previous prefix (e.g., `workload.googleapis.com`), you must include the full prefix in the metric name you define in your OpenTelemetry instrumentation. |
+
+#### Unsupported Features
+
+The following features of the `GoogleCloudMetricExporter` are not supported by the standard OTLP exporter:
+
+*   **Metric Descriptor Strategy (`setDescriptorStrategy`)**: OTLP exporters do not send metric descriptors separately. Metadata is handled automatically by the backend.
+*   **Custom Monitored Resource Mapping (`setMonitoredResourceDescription`)**: OTLP relies on standard OTel resources. GCP maps these to monitored resources automatically.
```
dashpole: We still support gcp.resource_type, right? Should we tell people to use that? Might be worth linking to https://docs.cloud.google.com/stackdriver/docs/reference/telemetry/v1.metrics#mr-mapping

MIGRATION.md
```diff
@@ -0,0 +1,220 @@
...
+| :--- | :--- | :--- |
+| `setProjectId(String)` | Use resource attribute: `gcp.project_id` | If using the `opentelemetry-gcp-auth-extension`, the project ID can be inferred from the credentials or the environment. |
+| `setCredentials(Credentials)` | Pass the bearer token as Authorization Header in the exporter | Handled automatically by `opentelemetry-gcp-auth-extension`. |
+| `setMetricServiceEndpoint(String)` | `otel.exporter.otlp.endpoint` / `OTEL_EXPORTER_OTLP_ENDPOINT` | Set it to `https://telemetry.googleapis.com` to send metrics to Google Cloud. |
+| `setDeadline(Duration)` | `otel.exporter.otlp.timeout` / `OTEL_EXPORTER_OTLP_TIMEOUT` | Default is 10 seconds. |
+| `setPrefix(String)` | N/A | The Telemetry API automatically prefixes metrics with `prometheus.googleapis.com/` by default and will be stored in Google Managed Prometheus with `prometheus_target` as the Monitored Resource. Custom prefixes are not directly supported via OTLP exporter configuration.<br>If you want your metric to keep appearing under your previous prefix (e.g., `workload.googleapis.com`), you must include the full prefix in the metric name you define in your OpenTelemetry instrumentation. |
```
dashpole: I don't think this is going to be acceptable.  We need to find a way for users to migrate to the OTLP exporter without breaking dashboards. There must be some way to do that, right? Is it possible with a view in java?  Or do we need to show users how to "wrap" their OTLP exporter to update metric names during the migration?

MIGRATION.md
aabmass, line 1: This seems duplicative of https://docs.cloud.google.com/trace/docs/migrate-to-otlp-endpoints

do we need both ?

README.md
```diff
@@ -1,5 +1,8 @@
 # Open-Telemetry Operations Exporters for Java
 
+> [!WARNING]
+> **DEPRECATION NOTICE**: This project and all its published artifacts are deprecated. No new features will be added, and this repository may be archived in the future.
```
aabmass: link to migration.md?

MIGRATION.md
psx95: Are you talking about the Logs Writer role?

dashpole: yes

MIGRATION.md
psx95: The trace one is, but this document has a little more information. Eventually we plan to update https://docs.cloud.google.com/trace/docs/migrate-to-otlp-endpoints and have that as single source of truth.

MIGRATION.md
```diff
@@ -0,0 +1,504 @@
...
+
+1.  In the Google Cloud console, on the project selector page, select or create a Google Cloud project.
+2.  Verify that billing is enabled for your Google Cloud project.
+3.  Enable the Telemetry, Cloud Logging, Cloud Monitoring, and Cloud Trace APIs.
+
+### Configure permissions
+
+Grant the following IAM roles to the service account that your application uses:
+
+*   **Cloud Telemetry Traces Writer** (`roles/telemetry.tracesWriter`)
```
dashpole: I thought the existing roles continue to work?

MIGRATION.md
```diff
@@ -0,0 +1,504 @@
...
+## Overview
+Google Cloud now supports native OTLP (OpenTelemetry Protocol) ingestion for Cloud Trace and Cloud Monitoring via the [Telemetry API](https://docs.cloud.google.com/stackdriver/docs/reference/telemetry/overview). This allows you to use the standard OpenTelemetry OTLP exporters for sending telemetry data to Google Cloud.
+
+## Before you begin
+
+Before you migrate your application to send trace data to the OTLP endpoint, ensure you have completed the following steps:
+
+### Enable billing and the Telemetry API
+
+1.  In the Google Cloud console, on the project selector page, select or create a Google Cloud project.
```
dashpole: People should already have these set up if they are migrating from our exporters.

MIGRATION.md
```diff
@@ -0,0 +1,504 @@
...
+
+    // Application-specific logic here
+
+    // Flush all buffered metrics on shutdown
+    openTelemetrySdk.getSdkMeterProvider().shutdown().join(10, TimeUnit.SECONDS);
+  }
+}
+```
+
+#### 4. Adding Attributes
```
dashpole, line 157: Not sure some of these points are neccessary. Why do we need to tell people how to add attributes?

dashpole (commented):
just nits

MIGRATION.md
psx95: No, the Telemetry API specific roles are required for successful exports.

MIGRATION.md
psx95: This is assuming the people migrating are not familiar with OpenTelemetry API and so the guide walks through some of the common usages.

It also serves as pre-cursor to resource attributes which is an important concept for a successful export.

MIGRATION.md
psx95: True, I will remove this section.

MIGRATION.md
psx95: Discussed offline, existing roles were indeed updated. I will remove this section. 
Thanks!

dashpole (approved):

PR #444, opened 2026-05-11
Add autoconfiguration auth extension example

psx95:


examples/autoconf-auth-extension/build.gradle
```diff
@@ -0,0 +1,48 @@
...
+}
+
+def autoconf_config = [
+	'-Dotel.exporter.otlp.endpoint=https://telemetry.googleapis.com',
+	'-Dotel.traces.exporter=otlp',
+	'-Dotel.metrics.exporter=otlp',
+	'-Dotel.logs.exporter=none',
+	'-Dotel.service.name=autoconf-auth-extension-example',
+	'-Dotel.exporter.otlp.protocol=http/protobuf',
+	'-Dgoogle.cloud.project=otel-test-sharmapranav',
```
dashpole: Maybe use GOOGLE_CLOUD_PROJECT here instead of your project id?

dashpole:
Does the auth extension set the gcp.project_id resource attribute?

examples/autoconf-auth-extension/build.gradle
psx95: Ah yes, forgot to remove after testing. Good catch! 👍🏻

psx95:
> Does the auth extension set the gcp.project_id resource attribute?

Yes

dashpole (approved):

